PR #66, opened 2017-06-27
Port to tidyselect

lionel-:
Depends on tidyverse/tidyselect#11.

It might make sense to rename `select_terms()` to `terms_select()` and make it take dots, so it's consistent with our general naming and UI schemes. Also, perhaps rename the `info` argument to `terms`. E.g. a function prefixed with `terms_` takes a `terms` object.

`vars_select()` now issue errors about "variables" rather than "columns". Let me know if you need to override any error message, I'll look into signalling typed conditions for those cases as well.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,8 @@ Imports:
     purrr,
     rlang (>= 0.1),
     gower,
-    RcppRoll
+    RcppRoll,
+    tidyselect
 Suggests:
     testthat,
     magrittr,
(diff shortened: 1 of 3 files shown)


topepo:
I was sure that we would need an altered version to handle the roles and types... I assume that this works because the `has_role` and `has_type` selectors work just like the ones in `dplyr`.

Thanks!

topepo:
Can you also look at the functions at and below [this line](https://github.com/lionel-/recipes/blob/f81a35e7165040c981105ad12e9d0f05d62b4151/R/selections.R#L350) to see if those functions need to be included in `recipes`? A few were just to get things working with the previous devel version of `dplyr`.

lionel-:
Yes those functions can probably be removed now.

Issue #62, opened 2017-06-21
step_intercept?


alexpghayes:
Thanks for this wonderful package! I'm currently using recipes to strip out some troublesome factor levels with `step_other` before passing a design matrix to `glm.fit`, and the package is making my life much better.

I was a little surprised that

```
data(okc)

okc %>% 
  recipe(age ~ height) %>% 
  step_center(height) %>% 
  prepare(okc) %>% 
  bake(okc)  
```

doesn't add an intercept column like I'd get if I just used `glm`. Are there plans for something like `step_intercept`, or is there an assumption that I should `cbind` on an intercept column outside of the recipe interface?

topepo:
I had not thought of that. Would you like to write one and submit a PR?

alexpghayes:
Sure. I'll try to find time before the end of the weekend. I'll include some changes to the documentation so nobody else gets surprised.

PR #75, opened 2017-07-13
implement and document step_intercept

alexpghayes:
Implementation of `step_intercept` that adds an intercept or constant valued column with role "predictor" re #62 .

Based off of `center.R`, although I don't know what the @keywords and @concept tags do so I omitted those in my documentation.

I brought my repo to up even with master and was having some odd troubles  with `Variable context not set` causing vignettes to not build and some tests to fail. Not sure why this was happening. Anyhow, none of the `step_intercept` code I added sets that off, so I'm assuming we're good.


alexpghayes:
i'm not sure what's broken here since my commits were after pulling and bringing my branch to even with `recipes/master`. more than willing to fix things, but may need a pointer in the right direction. are there any dependencies on dev versions of tidyselect or something like that?

topepo:
> `step_intercept` creates a _specification_ of a recipe step that will add an intercept or constant term in the first column of a data matrix. `step_intercept` has _predictor_ role so that it is by default called in the bake step. This means you should always call `step_intercept` last to avoid unintentional transformations such as centering, scaling, etc.

A better approach is to have another argument that is the name of the new column (maybe default to "intercept"). That would avoid the problem about the sequence of steps. Also, they can exclude the new column using the selectors so you don't have to worry about that. 

You can check the reason for the failure by running the tests yourself or look at the "Details" link above. This is failing because:
```r
> rec <- recipe(HHV ~ carbon + hydrogen + oxygen + nitrogen + sulfur,
+               data = biomass_tr)
> rec_trans <- recipe(HHV ~ ., data = biomass_tr[, -(1:2)])
> step_intercept(value = 2)
Error in rec$steps[[length(rec$steps) + 1]] <- object : 
  argument "recipe" is missing, with no default
Calls: step_intercept -> add_step
```
The first argument of ` step_intercept(value = 2)` should be a recipe so using something like:
```r
 rec_trans <- recipe(HHV ~ ., data = biomass_tr[, -(1:2)]) %>%
   step_intercept(value = 2)
```
would work.

alexpghayes:
Ah... forgot to compile the documentation changes. Also added a `name` argument to specify column name.

topepo:
Looks good. Thanks

Issue #74, opened 2017-07-10
Step for rectified linear unit


topepo:
Smooth and ramped, with an option to subtract a constant and another to reverse to get left hinge. document connection to MARS

topepo:
Added by @alexpghayes ! Thanks

PR #76, opened 2017-07-15
add step_relu

alexpghayes:
Saw #74 and realized it wasn't all that different from `step_intercept` so gave it a go.

Just saw the consistency tweaks to `step_intercept` and tried to use the same style here. Also tweaked the `step_intercept` tests following the changes to warnings/error messages.

Realizing I'm just submitting a PR out of the blue and not sure if this kind of thing helps or is a pain to manage. In general, are you open to PRs to address open issues? Is this the kind of thing where it's good to comment on the issue first to claim it, or something along those lines?


topepo:
This is great; it's pretty much what I was going to put in. Good MARS writeup. I'm glad someone understood my rambling. 

Can you make a few changes:

* sync to the master branch; I changed a few minor things for the intercept code, including the tests that you have in this PR. 
* don't replace the existing values (and thus default the role to "predictor")
* add a `prefix` option that defaults to "left_relu_" or "right_relu_" depending on the value of `reverse`? This is similar to `step_pca` to automatically naming the new columns. 
* add notes to `contributors.md` for your help

We need a print method too but I can take care of that.

alexpghayes:
Okay, think that does it. Still having trouble running R CMD Check locally, so some of those commits are single line changes or the like. I read about rebase but couldn't tell if it would make a mess so decided against for the time being.

alexpghayes:
We can avoid some unnecessary copying of data by using tidyeval to call `mutate` once, rather than `select` followed by `mutate_all` followed by assignment. However, my best current shot at this gives a `name ignored because a LHS was supplied` warning, which I presume isn't ideal.

```r
#' @importFrom dplyr select_vars tbl_vars mutate
#' @importFrom purrr map
#' @importFrom rlang expr sym
bake.step_relu <- function(object, newdata, ...) {
  col_names <- dplyr::select_vars(dplyr::tbl_vars(newdata), !!!object$terms)

  mut_exprs <- purrr::map(col_names, function(col) {
    new_col <- paste0(object$prefix, col)
    rlang::expr(
      !!rlang::sym(new_col) :=
        relu(!!rlang::sym(col), object$shift, object$reverse, object$smooth)
    )
  })

  suppressWarnings(dplyr::mutate(newdata, !!!mut_exprs))
}
```

For easy comparison, `bake.step_relu` without tidyeval:

```r
#' @importFrom dplyr select mutate_all select_vars tbl_vars funs
bake.step_relu <- function(object, newdata, ...) {
  col_names <- dplyr::select_vars(dplyr::tbl_vars(newdata), !!!object$terms)
  new_cols <- paste0(object$prefix, col_names)

  newdata[, new_cols] <- dplyr::mutate_all(
    dplyr::select(newdata, col_names),
    dplyr::funs(relu),
    object$shift,
    object$reverse,
    object$smooth)
  as_tibble(newdata)
}
```

topepo:
@lionel-, @hadley  Do you have any thoughts on @alexpghayes's versions of `bake.step_relu`?

R/relu.R
```diff
@@ -0,0 +1,144 @@
...
+#'   to true will instead calculate a smooth approximation to ReLu
+#'   according to \deqn{ln(1 + e^(x - c)} The \code{reverse} argument may
+#'   also be applied to this transformation.
+#'
+#' @section Connection to MARS:
+#'
+#' The rectified linear transformation is used in the Multivariate Adaptive
+#' Regression Splines as basis function to fit piecewise linear functions to
+#' data in a strategy similar to that employeed in tree based models. The
+#' transformation is a population choice as an activation function in many
```
lionel-: popular

R/relu.R
```diff
@@ -0,0 +1,144 @@
...
+    object$reverse,
+    object$smooth)
+  as_tibble(newdata)
+}
+
+relu <- function(x, shift = 0, reverse = FALSE, smooth = FALSE) {
+  if (!is.numeric(x))
+    stop("step_relu can only be applied to numeric data.", call. = FALSE)
+
+  if (reverse) shifted <- shift - x
```
lionel-: The branches should at least be placed on their own line and ideally wrapped in curly brackets.

R/relu.R
```diff
@@ -0,0 +1,144 @@
...
+      prefix = prefix
+    )
+  }
+
+prepare.step_relu <- function(x, training, info = NULL, ...) {
+  x$trained <- TRUE
+  x
+}
+
+#' @importFrom dplyr select mutate_all select_vars tbl_vars funs
```
lionel-: No need to import these functions if you are calling them with a namespace-qualifier as you do below

topepo:
@lionel- I was hoping to get feedback on this part:

> We can avoid some unnecessary copying of data by using tidyeval to call mutate once, rather than select followed by mutate_all followed by assignment. However, my best current shot at this gives a name ignored because a LHS was supplied warning, which I presume isn't ideal.

lionel-:
oops sorry.

The data in question is just a list of pointers (a data frame) so the current implementation should be fine in this regard. The issue is that `mutate_all()` is notoriously slow for wide data frames. Cf this performance comparison with a simple map: https://github.com/hadley/dplyr/files/867620/colwise-dplyr-purrr-nogroups.pdf. So if you have more than 50 terms in the design df, that might become an issue.

@alexpghayes this expression is not valid:

```r
rlang::expr(
  !! rlang::sym(new_col) :=
    relu(!! rlang::sym(col), object$shift, object$reverse, object$smooth)
)
```

That's because you're trying to assign a parameter name outside of a call. E.g. it's like you were trying to write `nm = expr` at the REPL. This would actually work but only because `=` is also an alias for the assignment operator `<-` when called at top-level as opposed to within a call. Since `:=` is not treated as an alias to `<-` you cannot call it at top-level.

I would just assign the names the usual way:

```r
make_relu_call <- function(col) {
  rlang::expr(relu(!! rlang::sym(col), object$shift, object$reverse, object$smooth))
})

names(col_names) <- paste0(object$prefix, col_names)
exprs <- purrr::map(col_names, make_relu_call)
```

The expression builder could also be:

```r
make_relu_call <- function(col) {
  rlang::lang("relu", rlang::sym(col), object$shift, object$reverse, object$smooth)
})
```

In the latter case all parameters are inlined within the call without having to `!!` everything.

I suggest using `#' @import rlang` to avoid these `rlang::` prefixes.

lionel-:
> In the latter case all parameters are inlined within the call without having to !! everything.

This should also produce a more informative call stack when an error is raised. The relevant calls will include the actual parameters rather than `object$param`.

R/relu.R
```diff
@@ -0,0 +1,144 @@
...
+#'   If \code{reverse} is true, then the transformation is reflected about the
+#'   y-axis, like so: \deqn{max(0, c - x)} Setting the \code{smooth} option
+#'   to true will instead calculate a smooth approximation to ReLu
+#'   according to \deqn{ln(1 + e^(x - c)} The \code{reverse} argument may
+#'   also be applied to this transformation.
+#'
+#' @section Connection to MARS:
+#'
+#' The rectified linear transformation is used in the Multivariate Adaptive
+#' Regression Splines as basis function to fit piecewise linear functions to
```
alexpghayes: as a

topepo: @alexpghayes We didn't get all of that =]

alexpghayes:
Think this should do it.

Thanks @lionel- for the feedback - didn't know about passing named lists to `mutate` but glad I do now. Also appreciate the comment about expressions, that makes sense. I'm greatly enjoying learning more about `tidyeval/rlang`.

topepo:
Thanks!

PR #103, opened 2017-12-01
Implement step_string2hashbucket()

kevinykuo:
This PR adds `step_string2hashbucket()` which allows encoding of character columns to hash buckets. It also adds `openssl` as an import to perform hashing.

Example usage:
```
set.seed(1)
df <- data.frame(V1 = sample(letters, 1000, replace = TRUE),
                 stringsAsFactors = FALSE)
recipe(~ V1, data = df) %>%
  step_string2hashbucket(V1, num_buckets = 10) %>%
  prep(df, stringsAsFactors = FALSE) %>%
  bake(df) %>%
  dplyr::count(V1)
# # A tibble: 9 x 2
#      V1     n
#   <int> <int>
# 1     0    37
# 2     1    83
# 3     3   265
# 4     4    35
# 5     5    89
# 6     6   113
# 7     7   119
# 8     8   140
# 9     9   119
```


topepo:
This was on my to-do list. Thanks! 

* Can we rename this to `step_featurehash` or similar? The name and some of the man file content might be unclear to someone whose never heard of this approach or know what a hash is. 

* I think that the resulting data type should be a factor (that includes a `prefix` option for naming or should prefix based on the original variable's name). Most people would make dummy variables out of this and that is easily facilitated by factors. 

* We should also accept factors as inputs. 

* I'm really against adding an external dependency. `caret` depends on `recipes` and adding an `imports` the requires a command line install would be disastrous since it gets download ~80K/month.  Can we use [`digest`](https://cran.r-project.org/web/packages/digest/index.html) package (in the same way that the [`FeatureHashing `](https://cran.r-project.org/web/packages/FeatureHashing/index.html) package does)? 

* Please make the integers 1-based.

kevinykuo:
Did some investigation and it seems like computing remainders on hashes isn't trivial. We'll likely want to create a package that does it cleanly and depend on that. Closing for now.

Issue #37, opened 2017-02-28
time variable lagging step


topepo:
For date variable inputs

PR #109, opened 2017-12-19
step_lag, step_naomit, detect_step and minor improvements

alexpghayes:
This PR includes the following:

### A new function `detect_step` to determine if a particular step occurs in a recipe

Modelling functions with a default X/y interface might have arguments like do_pca_first or include_intercept. `detect_step` makes it easier to wrap these interfaces with recipes interfaces.

### `step_lag` (#37)

Using this step induces NAs in some rows, which I presume is better than dropping the rows with holes induced by lagging.

Curious if it would be good to extend so that users can subtract off a fixed `lubridate::duration` from a column of dates.

### `step_naomit`

To complement imputation methods already in recipes and make it easier for people implementing models with recipes interfaces to separate data processing and model fitting (i.e. get rid of na.rm/na.action arguments in their fit/train methods)

Since `step_naomit` operates on rows rather than columns it probably needs a different name that reflects it isn't a step. Seems like a todo for the down the road?

### Minor improvements

- Print methods for step_intercept and step_relu
- Tidy method for step_relu
- Fix bug in step_relu when specifying all_predictors as selector, and add regression test


topepo:
Those look good! Don't forget to add to the NEWS file. 

`detect_step` could benefit from using `tidy.recipes`, which gives you a list of the current steps<sup>*</sup>

> Since `step_naomit` operates on rows rather than columns it probably needs a different name that reflects it isn't a step. 

I've broadened steps to include things that just affect rows (I'm writing a few now). `step_naomit` isn't a bad name but maybe `step_complete`? Either is fine with me. 

The plan is to also have a limited number of `dplyr` operations (like `filter`, `select`, and `mutate`) that can be used to do basic manipulations but I think that having this as a step is fine. 

> Curious if it would be good to extend so that users can subtract off a fixed `lubridate::duration` from a column of dates.

I'll need to look at `step_lag` for a while. Does it need to be focused on dates? I have data that are per-day. There is no date field but "positional lagging" by just shifting the row indices would be good. If that's too naive, can you add some examples that would differentiate the issues with dates vs non-dates? 

I really need to make progress on some workshop notes for the next few weeks so I don't know that I'll be able to examine at these in-depth for a bit. If you have specific questions though, just ping me here via `@topepo` and I will take a look. 

<sup>*</sup> A PR was just checked in that adds "checks" of the data (as opposed to steps) that can be added to the recipe. Also, there is a branch that I'll merge in soon that requires steps to have an argument (`skip`) that will allow the [operation to be skipped when using `bake`](https://github.com/topepo/recipes/blob/skips/vignettes/Skipping.Rmd). fyi...

topepo:
For `step_lag`, I'd want to be able to select many variables (that may not be dates) for lagging. 

Do we need a date? We could just tell people that it assumes that their rows are properly sorted. 

If we want to require a date field, then we should have a specific argument for it that people would wrap inside of `vars` and then let the `...` select the fields to lag. 

It would also be good to have the `lag` argument take multiple integers.

alexpghayes:
I think the default `dplyr::lag` behavior makes sense, and the commit I just added allows the user to lag any column. The current assumption is that the data is in some order the user wants before lagging.

The lag argument accepts a vector of lags, so multiple lags can be specified in a single step.

I couldn't find `tidy.recipes` for `detect_step`?

topepo:
No plural (`tidy.recipe`) =]

alexpghayes:
Ooooh gotcha. Just updated.

topepo:
Thanks

PR #135, opened 2018-03-19
recipe.formula() should return the resulting recipe object

DavisVaughan:
Just came across this when reading through the vignettes


topepo:
Thanks for catching this

Issue #141, opened 2018-04-04
Reference level for bin2factor


michaellevy:
Currently the level that replaces "1"s becomes the reference level. Is there a reason for that? My use case for this step is for outcomes, and there we definitely want whatever replaces "0" to be the reference level. If it were a predictor, I think you'd want the same... if you were getting a coefficient for it, you'd want the the change from no -> yes, but currently you'd get the opposite.

michaellevy:
Great. I'd like to work on this if you don't mind. I would think to do this by making the `levels` argument a named vector to specify which level replaces which integer and the first position becomes the reference level. Could do:

1. `levels = c("0" = "no", "1" = "yes")`, but that's awkward and inconsistent with `dplyr::rename` where `rename(newx = x)` means "replace x with newx". 
2. `levels = c(no = 0, yes = 1)` would require "if it's a named vector, the names are the levels, otherwise the entries are the levels", which is awkward.
3. Add an argument, maybe `ref_level`, that's one of the entries in `levels` and sets the reference.

Do you like any of those @topepo ?

topepo:
> I'd like to work on this if you don't mind.

I was hoping that you would say that. 😃

I would add some sort of logical that is something like `first_lev` (or better named). That way, the step can process a lot of columns that might have different levels.

topepo:
closed by #142

PR #142, opened 2018-04-06
add switch to make second level the reference in step_bin2factor

michaellevy:
First stab at addressing #141. I wrote the test differently than the others in the file because `table` orders by factor levels and I wanted to be explicit about what's being tested here.

### DIFF
--- a/R/bin2factor.R
+++ b/R/bin2factor.R
@@ -13,6 +13,8 @@
 #' @param levels A length 2 character string that indicate the
 #'  factor levels for the 1's (in the first position) and the zeros
 #'  (second)
+#' @param ref_first Logical. Should the first level, which replaces
+#' 1's, be the factor reference level?
 #' @param columns A vector with the selected variable names. This
 #'  is `NULL` until computed by [prep.recipe()].
 #' @return An updated version of `recipe` with the new step
@@ -51,6 +53,7 @@ step_bin2factor <-
            role = NA,
            trained = FALSE,
            levels = c("yes", "no"),
+           ref_first = TRUE,
            columns = NULL,
            skip = FALSE) {
     if (length(levels) != 2 | !is.character(levels))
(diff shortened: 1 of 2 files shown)


michaellevy:
Suggest making the default of the new argument FALSE. Binomial glms, e.g., assume the first level is the negative class, which leads to this backwardness with the current default.

``` r
suppressPackageStartupMessages({
  library(tidyverse)
  library(recipes)
})
d <- 
  tibble(y = rbinom(50, 1, .5),
         x1 = rnorm(50, mean = y),
         x2 = rnorm(50, mean = y))
d <- 
  recipe(d) %>% 
  step_bin2factor(y) %>% 
  prep(d) %>% 
  bake(d)
m <- glm(y ~ ., d, family = "binomial")
d$predicted_prob <- predict(m, d, type = "response")
ggplot(d, aes(x = y, y = predicted_prob)) + 
  geom_boxplot()
```



Created on 2018-04-06 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

topepo:
> Suggest making the default of the new argument FALSE.

No, I'm sticking to my guns on this one. Sorry about the rant. 

This convention is old and misguided since it is based on encoding binary categorical data as 0/1. If people forgot about this and were asked "which factor level is generally more important?," nobody would say "the last one". 

So in my quixotic efforts to change established human behavior, I'm going to say that we should treat categorical data as qualitative and forget about how it is eventually encoded (i.e., please make the default TRUE). 

`</rant>`

michaellevy:
Quixotic ranting is good. I'll save you my counter-rant because I see that this is baked into caret and so would be a massive change. The default is TRUE.

topepo:
:+1:

Can you add a unit test for the option and make a note in the NEWS.md file under "Other Changes"? Feel free to link to the PR or issue and use "(contributed by...)" or similar.

michaellevy:
Sorry, I missed that last ask. Thanks for letting me add this!

PR #145, opened 2018-04-16
Feature `one_hot` argument for `step_dummy()`

DavisVaughan:
Since I sadly deleted the PR #136 branch with `one_hot`, I could not reopen it and have reimplemented it here, incorporating your comments from there. 

I have not removed the `contrasts` argument, as I feel that is a separate feature.

### DIFF
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -78,6 +78,20 @@
 #' unique(okc$diet)
 #' grep("^diet", names(dummy_data), value = TRUE)
 #'
+#' # Obtain the full set of dummy variables using `one_hot` option
+#' rec %>%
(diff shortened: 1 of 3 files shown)


DavisVaughan:
Regarding the `contrasts` argument, trying the code below with all 5 of the different unordered contrast types set as global options results in behavior mimicking a fallback to using the default, `contr.treatment`, with `one_hot = TRUE`. This probably makes sense right?

```r
data(okc)
okc <- okc[complete.cases(okc),]

rec <- recipe(~ diet + age + height, data = okc)

rec %>% 
  step_dummy(diet, one_hot = TRUE) %>%
  prep(training = okc) %>%
  bake(newdata = okc)
```

PR #149, opened 2018-05-08
Do not raise error in convert_matrix() if data frame contains logical columns

jrnold:
Currently, `convert_matrix` will raise an error if any column is logical. This is annoying because often design matrices contain indicator variables which may be stored as numeric or logical vectors.

This commit updates `convert_matrix` to not throw an error if there are any logical columns, and to convert logical columns to numeric prior to running the `*model.matrix()` functions.


jrnold:
I guess I considered logical to be a special case because it is an atomic type, and it occurs quite naturally in design matrices. I wouldn't be surprised if any other classes weren't coerced into numeric. I ran into this a bunch of times, and had to coerce logical vectors into numeric multiple times before sending the data frame to recipes.  Though, since model matrix treats logical variables like factor variables rather than numeric variables, I shouldn't have been.

But your position completely makes sense. However, now I don't understand what you'd want.  Since `model.matrix()` or `sparse.model.matrix()` is used for coercing the data frame into a matrix, it seems that ensuring that all variables are numeric is the only safe thing to do.

jrnold:
Reading back over your original comments, is your idea to coerce the data frame to a matrix without going through model matrix?

jrnold:
I wonder if there is a reason to use `model.matrix` at all? If all the conversions in the design matrix are occurring in steps, what is left for `model.matrix` to do? And more importantly, would you want `model.matrix` to do some things implicitly that weren't specified in steps? It seems that it should just use `as.matrix` or the equivalent for sparse matrices.

topepo:
> I wonder if there is a reason to use `model.matrix` at all?

I can't think of the reason that I used it. It may end up being a problem if someone doesn't convert all of their factors to dummy variables (since it will do that for them). 

`as.matrix` does seem like the correct way to go with the added check that all of the columns are a common type.

jrnold:
> I can't think of the reason that I used it.

Maybe it was doing some of the lifting before more steps were introduced? It may be good to have all the attributes associated with the matrix output by `model.matrix()`, if some downstream function expects it, but I think most of those are optional since it returns a `matrix` object and won't always create various attributes.

> It may end up being a problem if someone doesn't convert all of their factors to dummy variables (since it will do that for them).

That would be a bug because wouldn't you want them to use a step to do this? In any case, the existing `is.numeric` check would raise an error if there were any factors in the data frame.

> `as.matrix` does seem like the correct way to go with the added check that all of the columns are a common type

I can do that.  Do you want the matrix to be a common type or to always be numeric? If recipes is a preprocessor for design matrices I'd expect it to be numeric, but if it is a more general preprocessor, I could see other outputs.

topepo:
> Do you want the matrix to be a common type or to always be numeric? If recipes is a preprocessor for design matrices I'd expect it to be numeric, but if it is a more general preprocessor, I could see other outputs.

Common type. There are some text-based steps that we'll be adding in the future.

topepo:
Is this PR where we want it? I'm doing a CRAN release in the next few days.

jrnold:
This may be one that you have to do since it more about your vision. Here was my take on it. Instead of trying to infer what type of matrix the user wants make the user specify the type of the matrix. 

- Add "character" option to `composition` argument of `bake`. This returns a character matrix.
- Add `type` argument to `convert_matrix` function. This takes values "numeric" or "character" (
- Since `model.matrix` and `sparse.model.matrix` are used for converting to numeric matrices allow factor and logical variables. With sparse matrices, this avoids ever creating dense columns of indicator variables.

juliasilge:
Thank you so much for your helpful discussion and input @jrnold! 🙌 We are cleaning up older issues and PRs, but appreciate your input here.

PR #150, opened 2018-05-10
Indicate necessary version of purr (0.2.3)

karissawhiting:


karissawhiting:
Current version of recipes requires `map_dfr()` which was introduced in purrr 0.2.3

Issue #148, opened 2018-05-02
Adding a signed_log


EdwinTh:
I personally find this very helpful for taking a log on variables that contain negative values. It is defined as 

    ifelse(abs(x) < 1, 0, sign(x) * log(abs(x))

I have the code already for this, as it is in the [eRum presentation](https://github.com/EdwinTh/recipe_for_recipes/blob/master/eRum.Rmd). I could either do a PR for a new step, or I could modify the existing `step_log`. My preference would be the former, but please tell me what is yours.

artichaud1:
Sorry for hijacking, but I wonder: is the fact that there isn't a general `step_func` on purpose ? (esp. for cases like this one where the `prep` function would actually do nothing). If so I would be curious to know the rationale.

topepo:
That might be a bit of a nightmare to support since there would probably be a lot of things that could go wrong. I will be implementing a few `dplyr` verbs (`filter` and `mutate`) that can be used in steps. would that cover what you want to do?

topepo:
> I could modify the existing `step_log`

That is my preference.

EdwinTh:
Ok will do

PR #158, opened 2018-06-12
Adding signed arg to log, as disscussed in #148

EdwinTh:
R CMD check failed because something seems wrong in the DESCRIPTION file. I didn't touch it, so I don't think I did that.


topepo:
Thanks. Right under the gun for a CRAN release

PR #170, opened 2018-07-14
Fix typo

lawwu:
I believe this is what you mean to say?


topepo:
Thanks for finding these.

PR #200, opened 2018-09-27
documentation for check_type

### LINKED ISSUE #194: Improve docs for check_type
I know this is _marked_ as internal, but it's exported andincluded in the docs, yet there's no notation for either of it's args or explanation of what it does/how to use it. It's used all over the place, and seems like something one would want to add to their own steps, so it should probably be more formally documented, and possible discussed in the Custom Steps vignette.

EmilHvitfeldt:
hopefully this should be able to resolve #194


topepo:
Thanks

PR #205, opened 2018-09-29
Include example using imp_vars

EmilHvitfeldt:
I found it a little puzzling that imp_vars was included in the documentation but wasn't used in the examples. So I added a example showing how to use `imp_vars`.


topepo:
Thanks

Issue #208, opened 2018-10-02
Name collusions


EmilHvitfeldt:
Using a step that creates new variables to error or at least throw a warning when the created variables matched existing variable names.
In the following case the 3 principal component variables end up being named `PC11`, `PC2` and `PC3`.

``` r
library(recipes)
library(dplyr)
data <- USArrests %>%
  mutate(PC1 = "collusion")  

rec <- recipe( ~ ., data = data)
pca_trans <- rec %>%
  step_center(all_numeric()) %>%
  step_scale(all_numeric()) %>%
  step_pca(all_numeric(), num = 3)

pca_estimates <- prep(pca_trans, training = data)
bake(pca_estimates, data)
#> # A tibble: 50 x 4
#>    PC1          PC11     PC2     PC3
#>    <fct>       <dbl>   <dbl>   <dbl>
#>  1 collusion -0.976   1.12   -0.440 
#>  2 collusion -1.93    1.06    2.02  
#>  3 collusion -1.75   -0.738   0.0542
#>  4 collusion  0.140   1.11    0.113 
#>  5 collusion -2.50   -1.53    0.593 
#>  6 collusion -1.50   -0.978   1.08  
#>  7 collusion  1.34   -1.08   -0.637 
#>  8 collusion -0.0472 -0.322  -0.711 
#>  9 collusion -2.98    0.0388 -0.571 
#> 10 collusion -1.62    1.27   -0.339 
#> # ... with 40 more rows
```

<sup>Created on 2018-10-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

PR #211, opened 2018-10-03
Throw error with name collision

EmilHvitfeldt:
In reference to #208, I have been thinking about it most of the day. To me it makes most sense to give an error instead of a warning. If we were to throw a warning you would have to recursively search for names that wouldn't collide, and that search space would be more and more obscure.

``` r
library(recipes)
library(dplyr)
data <- USArrests %>%
  mutate(PC1 = "collusion")  

rec <- recipe( ~ ., data = data)
pca_trans <- rec %>%
  step_center(all_numeric()) %>%
  step_scale(all_numeric()) %>%
  step_pca(all_numeric(), num = 3)

pca_estimates <- prep(pca_trans, training = data)
#> Error: Name collision occured in `step_pca()`. The following variable names already exists: PC1.
```

<sup>Created on 2018-10-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>


R/misc.R
```diff
@@ -457,3 +457,13 @@ simple_terms <- function(x, ...) {
   res
 }
 
+check_name <- function(x, y, name) {
```
topepo: Maybe pass in the step object and use the first class to write the warning to avoid typos

topepo:
Agreed on warning v  error. 

As an FYI we've been [talking about this in general](https://github.com/tidyverse/tibble/pull/469) in the tidyverse group so there will probably be a better way of handling this at the tibble-level in the near future.  

One change: can you make the specific column name changes via a function instead of repeated similar code? Maybe expand the checking code to include repair.

topepo:
> Would you want it to repair instead of erroring?

My mistake. Error is fine (we'd wait for tibble changes for repair). I just want to removed repeated code.

EmilHvitfeldt:
> My mistake.
No problem!

> Error is fine (we'd wait for tibble changes for repair).
Sure thing!

> I just want to removed repeated code.
Have I done that now? or have I overlooked something?

topepo:
I was thinking that you have these repeated lines in a bunch of places:
```r
  newname <- names0(ncol(comps), object$prefix)
  check_name(colnames(newdata), newname, object)
  colnames(comps) <- newname
```
and they could be made into a function that can be called. If we ever need to change this code, we do it in one place.

EmilHvitfeldt:
Okay I managed to remove some lines!

Unfortunately the naming scheme in `lag`, `depth`, `classdist` and `relu` doesn't use `prefix` and `names2` so I had to add a couple of arguments in check_name.

PR #224, opened 2018-10-13
Custom transformation step

smaakage85:
Hi Max

Here is my first of two (higher order) recipes steps. It gives the user the opportunity to do almost whatever transformation he or she wants to the data without having to write dedicated step functions from scratch each time. We already use the step a lot in our model development workflow at the Danish Tax Authorities. I hope, that you will also find it useful.  Please let me hear, what you think.

I was wondering, if "step_higher_order" would be a better name for it, but I don't have a strong opinion on that.

Best,
Lars


topepo:
It looks great but I think that it should live in its own package (like `embed` and the forthcoming [`textrecipes`](https://github.com/EmilHvitfeldt/textrecipes)).

smaakage85:
Hi @topepo. OK, I will put it in a separate package. :)

smaakage85:
When do you think, you will release the next version of recipes? Best Lars
________________________________
Fra: Max Kuhn <notifications@github.com>
Sendt: 14. oktober 2018 22:16:19
Til: tidymodels/recipes
Cc: Lars Kjeldgaard; Author
Emne: Re: [tidymodels/recipes] Custom transformation step (#224)


It looks great but I think that it should live in its own package (like embed and the forthcoming textrecipes<https://github.com/EmilHvitfeldt/textrecipes>).

topepo:
In about 2-3 weeks

PR #225, opened 2018-10-15
Allow list columns when computing the number of complete records

topepo:


topepo:
Changes for EmilHvitfeldt/textrecipes#12

PR #235, opened 2018-10-28
dplyr recipes steps

topepo:


R/mutate.R
```diff
@@ -0,0 +1,162 @@
...
+      role = role, 
+      input = inputs,
+      skip = skip,
+      id = id
+    )
+  )
+}
+
+step_mutate_new <- 
+  function(terms, role, trained, input, skip, id) {
```
lionel-: It would be clearer if `input` was plural. Perhaps `dots` or `exprs` would be even clearer?

R/mutate.R
```diff
@@ -0,0 +1,162 @@
...
+  role = "predictor", 
+  trained = FALSE, 
+  input = NULL,
+  skip = FALSE,
+  id = rand_id("mutate")
+) {
+  
+  inputs <- enquos(...)
+  if (is_empty(inputs)) 
+    stop("Please supply at least one name-value pair.", call. = FALSE)
```
lionel-: Is it necessary to fail on empty inputs? That's inconsistent with dplyr and this makes the function less composable and reusable in other functions.

R/mutate.R
```diff
@@ -0,0 +1,162 @@
...
+  if (any(names(inputs) == ""))
```
lionel-: Have you considered `enquos(..., .named = TRUE)`? That gives default names to expressions which is useful for quick interactive use.

R/mutate.R
```diff
@@ -0,0 +1,162 @@
...
+
+#' @importFrom rlang quo_get_expr expr_text
+#' @importFrom purrr map map_chr
+#' @importFrom dplyr tibble
+#' @rdname step_mutate
+#' @param x A `step_mutate` object
+#' @export
+tidy.step_mutate <- function(x, ...) {
+  var_expr <- map(x$input, quo_get_expr)
+  var_expr <- map_chr(var_expr, expr_text, width = options()$width, nlines = 1)
```
lionel-: Should be `quo_text(x)` or `expr_text(quo_squash(x))`. Unwrapping the quosure is not sufficient, there might be nested quosures in the expression. The rlang deparsing API is not great, in particular the duplication between `expr_` and `quo_` functions is a bit silly. We'll review it for rlang 0.4.0.

R/mutate.R
```diff
@@ -0,0 +1,162 @@
...
+#' @inheritParams step_center
+#' @param ... Name-value pairs of expressions. See [dplyr::mutate()].
+#' @param role For model terms created by this step, what analysis
+#'  role should they be assigned? By default, the function assumes
+#'  that the new dimension columns created by the original variables
+#'  will be used as predictors in a model.
+#' @param input Quosure(s) of `...`.
+#' @return An updated version of `recipe` with the new step
+#'  added to the sequence of existing steps (if any). For the
+#'  `tidy` method, a tibble with columns `values` which
```
lionel-: Maybe mention that the expressions in `values` (strange name for that column, but I guess that's consistent with other tidy methods?) are not reparsable because the quosures have been squashed, and you couldn't represent the environment as string anyway. These deparsed expressions are purely indicative.

R/mutate.R
topepo: done

R/mutate.R
topepo: It bothers me that this could happen but is a good idea.

PR #244, opened 2018-11-09
Fix/issue 206

DavisVaughan:
@topepo you mentioned that we could not use `as.matrix()` because of factors, but when `composition = "matrix"` is used, it is enforced that all columns are already numeric (taken care of by `step_dummy()` or something). So we should be fine.

The only difference between `as.matrix()` and `model.matrix(~ . + 0)` is then that the latter keeps the default row names of the data frame (integers 1, 2, 3) and adds them to the matrix, but `as.matrix()` does not keep them. Both keep row names if the user had supplied them to the data frame. I like this behavior better personally.


topepo:
> The only difference between `as.matrix()` and `model.matrix(~ . + 0)` is then that the latter keeps the default row names of the data frame (integers 1, 2, 3) and adds them to the matrix, but `as.matrix()` does not keep them. Both keep row names if the user had supplied them to the data frame. I like this behavior better personally.

That sounds fine to me. Ship it!

PR #245, opened 2018-11-12
Feature) Multiple roles, Issue 221

### LINKED ISSUE #221: roles and possible type to list columns
In the spirit of breaking changes for the next version...  ☹️

For example:

``` r
library(recipes)
#> Loading required package: dplyr
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
#> 
#> Attaching package: 'recipes'
#> The following object is masked from 'package:stats':
#> 
#>     step
recipe(Sepal.Length ~ ., data = iris) %>%
  summary()
#> # A tibble: 5 x 4
#>   variable     type    role      source  
#>   <chr>        <chr>   <chr>     <chr>   
#> 1 Sepal.Width  numeric predictor original
#> 2 Petal.Length numeric predictor original
#> 3 Petal.Width  numeric predictor original
#> 4 Species      nominal predictor original
#> 5 Sepal.Length numeric outcome   original
```

Created on 2018-10-08 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)

There are certainly occasions where a column can play more than one role. 

We also might want to have more granular `type` attributes. For example, should a date field have types of "numeric" and "date"?

This would affect the functionality of `add_role` and we might need an additional `change_role` since the former should append and the latter should reset.

DavisVaughan:
This PR adds support for multiple roles (and types) for a single variable. Main changes are:

* `add_role()` now only adds _new_ roles for existing variables to the recipe. I think this is more in line with its name too.
* `update_role()` alters an existing role with a new role name. If multiple roles are present for a single variable, it updates all of them to the `new_role` unless the user specifies `old_role`, in which case it will only alter that one.
* The internals of `current_info()` has changed a bit so that it holds `vars` (same as before) and `data` which is a list the same length as `vars` of tibbles containing the type/role information for each variable. This is a good way to compactly store multiple roles for a single variable. It was also motivated by `tidyselect` requiring that the input vars must be unique, which this format supplies.

This allows for a couple cool things, one of which is shown below. Note that `all_predictors()` would have selected `Sepal.Width` and `Petal.Length`, but the secondary role prevented that.

``` r
library(recipes)
recipe(Species ~ ., iris) %>%
  add_role(Sepal.Width, Petal.Length, new_role = "dontcenterme") %>%
  step_center(all_predictors(), -has_role("dontcenterme")) %>%
  prep(training = iris, retain = TRUE) %>%
  juice
#> # A tibble: 150 x 5
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species
#>           <dbl>       <dbl>        <dbl>       <dbl> <fct>  
#>  1       -0.743         3.5          1.4      -0.999 setosa 
#>  2       -0.943         3            1.4      -0.999 setosa 
#>  3       -1.14          3.2          1.3      -0.999 setosa 
#>  4       -1.24          3.1          1.5      -0.999 setosa 
#>  5       -0.843         3.6          1.4      -0.999 setosa 
#>  6       -0.443         3.9          1.7      -0.799 setosa 
#>  7       -1.24          3.4          1.4      -0.899 setosa 
#>  8       -0.843         3.4          1.5      -0.999 setosa 
#>  9       -1.44          2.9          1.4      -0.999 setosa 
#> 10       -0.943         3.1          1.5      -1.10  setosa 
#> # ... with 140 more rows
```

I've added as many tests as I can think of, but if you have any ideas on more ways to try and confuse this I'd like to hear them!

Closes #221


topepo:
Can you add a note to `?summary.recipe` to mention that there can be multiple rows for the same variable now? Is there anywhere else that we should talk about this?

tests/testthat/test_roles.R
```diff
@@ -27,31 +27,284 @@ test_that('changing roles', {
...
+      prep(training = biomass, retain = TRUE) %>%
+      juice() %>%
+      ncol(),
+
+    8
+  )
+
+})
+
+test_that("type selectors can be combined", {
```
topepo: I'm getting a warning during `test()` that there is no `expect_*` here.

DavisVaughan: Oh whoops ill fix that

topepo: I can add it to my branch

Issue #177, opened 2018-07-30
harmonize parameter names with `dials`


topepo:
It would be helpful to slowly harmonize recipe arguments that could be tuning parameters:

 - [x] `K` in `step_knnimpute` to `neighbors`. 
 - [x] `df` and `degree` in the spline steps; these need to be promoted to main args to the function
 - [x] standardize components in `step_pca`, `step_pls`, `step_isomap`, `step_kpca`, `step_ica` to use  `num_comp` (and `svm_sigma` for kPCA)
 - [x] add a `dials` parameter for threshold on [0,1]
 - [x] `step_bagimpute`: move `nbagg` out into a main option and name it `trees`. 
 - [x] `nunique` to `num_unique` in BC and TJ transformations
 - [x] `step_isomap` promote `knn` to main argument and rename

See if this can be done without breaking backward compatibility (maybe using the `...` to catch the old arguments and issue a warning).

alexpghayes:
It's ambitious, but I think `tidymodels/principles` (will move to tidymodels when I get a second) should define a glossary of standard concepts and naming schemes to use across the tidymodels packages.  Broom is currently in the process of standardizing terms as well, and will then tests argument names and output columns names against glossaries defined in [modeltests](https://github.com/alexpghayes/modeltests).

In tidymodels/broom#452 I've started some brainstorming about concepts that it would be good to provide consistent language for.

1. Would you be willing to add the names that you settle on to the `Naming Things` section of `principles`?
2. If I wrote tests to enforce consistent names, would you use them?

PR #267, opened 2018-12-04
Added condition to `nn_ind` to capture K

GoldbergData:
I added a condition to capture old `K` arguments supplied to step_knnimpute(). I encountered the error "$ Operator is invalid for atomic vectors" since object$neighbors did not exist for a recipe I prepped prior to the implementation of https://github.com/tidymodels/recipes/issues/177. It would be a hassle to completely retrain the data. I was able to informally fix the issue with: model$recipe$steps[[2]]$neighbors <- 5, but this proposed change makes it more formal.


topepo:
Thanks

PR #304, opened 2019-03-19
Fixed Issue 301 - news and unit test included

meztez:
in prep.step_dummy, the terms object has an embedded environment. Setting it NULL results in significant reduction in serialized recipes size when using save, saveRDS or other save methods.


N1KG:
sharp af

topepo:
Can you add an additional test to save a prepped recipe to a tmp directory (see `?tempfile`), reload, and bake some data? Please use `save()` and `saveRDS()` to test.

topepo:
Thanks!

PR #325, opened 2019-04-29
Add option to return a tibble containing matrices

jkennel:
Group regressors by **role** into matrices.  These matrices are then put into a tibble. This allows one to specify the matrix column of the tibble in model calls.  For example, after baking a recipe one can now say lm(outcome~lag_variable, rec) where rec is a tibble containing matrices where outcome is a matrix of one or more columns and lag_variable is a matrix column of one more than one columns. This also simplifies **predict** when **type = 'terms'** when we want our regressor output grouped (ie each regressor matrix yields one predict column).

Additional changes:
- added a couple of tests
- fixed two typos in Multivariate Analysis using Partial Least Squares vignette

### DIFF
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -604,6 +604,8 @@ bake.recipe <- function(object, new_data = NULL, ..., composition = "tibble") {
     new_data <- base::as.data.frame(new_data)
   } else if (composition == 'tibble_of_matrix') {
     new_data <- convert_tibble_of_matrices(new_data, object$term_info)
+  } else if (composition == 'data.frame_of_matrix') {
+    new_data <- convert_df_of_matrices(new_data, object$term_info)
   }
 
   new_data
@@ -775,11 +777,13 @@ juice <- function(object, ..., composition = "tibble") {
     new_data <- base::as.data.frame(new_data)
   } else if (composition == 'tibble_of_matrix') {
     new_data <- convert_tibble_of_matrices(new_data, object$term_info)
+  } else if (composition == 'data.frame_of_matrix') {
+    new_data <- convert_df_of_matrices(new_data, object$term_info)
   }
 
   new_data
 }
 
-formats <- c("tibble", "dgCMatrix", "matrix", "data.frame", 'tibble_of_matrix')
+formats <- c("tibble", "dgCMatrix", "matrix", "data.frame", 'tibble_of_matrix', 'data.frame_of_matrix')
 
 utils::globalVariables(c("number"))
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -602,6 +602,8 @@ bake.recipe <- function(object, new_data = NULL, ..., composition = "tibble") {
     new_data <- convert_matrix(new_data, sparse = FALSE)
   } else if (composition == "data.frame") {
     new_data <- base::as.data.frame(new_data)
+  } else if (composition == 'tibble_of_matrix') {
+    new_data <- convert_tibble_of_matrices(new_data, object$term_info)
   }
 
   new_data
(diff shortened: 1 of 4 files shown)


topepo:
Thanks for the PR. 

Could you help us understand the general use case more? I get the _tactics_ of what you are doing but some idea of the _strategy_ that you want to enable would help a lot (with some examples). 

My first thoughts are: 

 * how often will this be used?
 * how well would it be understood? 
 * are there a lot of ways that someone might misuse it? Especially of they specify the roles and variables incorrectly?  
 * Would it be better as a step?

jkennel:
# Thoughts on use case

Using matrices in the formula specification to a model allows **predict.lm** **type = 'terms'** to group related terms.  I work with time series and often have terms that should be grouped.  One such use case is when fitting to a shifted sin curve by using sin and cos terms.  We will fit two harmonics using four regressors, however, in my prediction I would prefer two prediction terms of different frequencies as the regressor terms (sin and cos) can be easier to interpret when taken together: 
  
`
x1 <- seq(0, 8 * pi, length.out = 100)
x2 <- seq(0, 4 * pi, length.out = 100)

y <- sin(x1 + pi/6) + sin(x2 + pi/8) + rnorm(length(x1), mean = 0, sd = 0.1)

df <- data.frame(y = y,
                 s_term1 = sin(x1), c_term1 = cos(x1),
                 s_term2 = sin(x2), c_term2 = cos(x2))
fit_1 <- lm(y~s_term1 + c_term1 +s_term2 + c_term2 - 1, df)


library(tibble)
df <- as.tibble(list(y = as.matrix(y),
                     x_comb1 = cbind(s_term = sin(x1), c_term = cos(x1)),
                     x_comb2 = cbind(s_term = sin(x2), c_term = cos(x2))))

fit_2 <- lm(y~x_comb1+x_comb2-1, df)

summary(fit_1)
summary(fit_2)

all.equal(coef(fit_1), coef(fit_2), check.attributes = FALSE)

head(predict(fit_1, type = 'terms'))
head(predict(fit_2, type = 'terms'))

`

Another example (but with **data.frames**) is evident in the vignette on Multivariate Analysis using Partial Least Squares.  Here juice is run twice and there is still post-processing in the function **get_var_explained**

`
y_mat <- juice(recipe, composition = "matrix", all_outcomes())
x_mat <- juice(recipe, composition = "matrix", all_predictors())

pls_format <- data.frame(
  endpoints = I(y_mat),
  measurements = I(x_mat)
)
`

If you have the option to output a **tibble** of matrices (or **data.frame** of matrices which I just added to the PR) this portion of code is not really needed as you can directly output pls_format from **juice** to use in your model.  I don't know how often this is used, but the fact that one of the examples (and one of the only ones that goes on to fit a model) in the vignettes would benefit from it suggests that it might be somewhat common. 

I could see it being applicable to lagged processes as well, where we might be interested in the sum of the lagged terms more than each lag separately.  I'm thinking of this in relation to the downstream output of **predict**, and the ubiquitous **predict.lm** in particular.

Spline fits (bs, ns, ...) have multiple terms and are commonly used.  On their own the predictions from each term might not be that useful (unless additional wrangling is performed).  However, as a whole (summed) form they do. This is a efficient way to get the combined spline component from predict. 

It can simplify model specification, for examples where you have groups that you would like to add or not.   However, if you want to remove a specific regressor in a group the **tibble_of_matrices** makes it more difficult (ideally, I would prefer the removal/exclusion/update done using a recipe).

1) **how often will this be used?** I can't really say but I think for signal decomposition it could be used quite a bit, and it depends on what models people are using
2) **how well would it be understood?** The data and actual fits are the same, we just group to achieve more efficient downstream predictions and model formulations.
3) **are there a lot of ways that someone might misuse it? Especially if they specify the roles and variables incorrectly?** I don't know how people are using roles and have only used **recipes** a short while.   In fact, it still isn't clear to me how/why they are used after reading the vignette.   My assumption is that they provide excellent information for grouping, but if you had other ideas in mind for roles I could see that we might not want to overload them with this.  My thinking was that I might have roles called lagged, harmonic, trend, other, outcomes, ...  Because the entry for the name of the **composition** is more complicated (**tibble_of_matrix**), it may deter some from using it without actually desiring the feature. 
4) Would it be better as a step?  I like it as a step, but I think the way the code is currently implemented it seemed most straightforward and possibly more efficient to put it in bake and juice.

juliasilge:
Thank you so much for this PR @jkennel! Although we haven't implemented anything exactly like what you have here, we appreciate your input and thought, and as the recipes ecosystem has developed, the idea of nested data has in fact been something we have worked on, especially in [textrecipes](https://github.com/tidymodels/textrecipes).

PR #360, opened 2019-07-08
ensure that prepper sets prep fresh argument to TRUE

### LINKED ISSUE #336: prepper should always use fresh = TRUE
(in reference to tidymodels/aml-training#1)

If it is false, and the recipe has already been prepped, the same recipe is inserted for all resamples.

smwindecker:
Closes #336


DavisVaughan:
@topepo we walked through this one together. Look good to you?

DavisVaughan:
Thanks!

PR #361, opened 2019-07-08
Modify threshold argument for step_other

### LINKED ISSUE #289: `n` argument for `step_other`
Would like if `threshold` argument within `step_other()` could be specified by an integer (rather than just by proportion), or had a separate `n` argument to specify the minimum sample size before it is collapsed into the "other" category.  
E.g. to specify minimum sample size of 30:
```r
recipe(price ~ clarity + color + carat, data = diamonds) %>% 
  step_other(all_nominal(), n = 30)
```

I'm curious what the best way of doing this is currently? ([Link](https://gist.github.com/brshallo/a25cc71fc66cbdee4e8ce51b1f74d5ca) to hack for specifying consistent sample size to `step_other` across datasets of different sizes.)

agila5:
Modify threshold argument of step_other in two ways: 

1) the prop argument is now named threshold; 
2) if threshold is an integer greater or equal than 1 then it specifies the minimum sample size before the factor levels are collapsed into the "other" category. If threshold is exactly equal to 1 then it removes the unused level of the factor. 

 Should close #289.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -18,6 +18,9 @@ Release driven by changes in `rlang`.
  
 ## Other Changes:
 
+* If `threshold` argument of `step_other` is greater than one then it specifies the minimum sample size before the levels of the factor are collapsed into the "other" category. [#289](https://github.com/tidymodels/recipes/issues/289)
+
+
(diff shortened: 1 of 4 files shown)


DavisVaughan:
Thanks @agila5! Great work!

agila5:
You're welcome :)

PR #375, opened 2019-08-22
Issue 111

### LINKED ISSUE #111: check_classes
Description:
`prep.check_classes` will learn the classes of each of the colums.
`bake.check_classes` breaks when:
- column of which the class is learned is missing from `newdata`
- column is a of a different class than was learned in `prep`

Choices to make:
- do we consider `integer` and `numeric` as the same class? (I would say yes)
- do we consider `factor` and `character` as the same class? (I would say maybe add an argument for this that defaults to `TRUE`)
- do we allow for date time variables? (I would say yes).

EdwinTh:
With the recent changes made to the package, this code now works (earlier character - factor problems no longer occurs), so it can be merged. Will close #111 

`devtools::check()` give the following warnings and note, which don't seem related to the changes made:

❯ checking for missing documentation entries ... WARNING
  Undocumented code objects:
    ‘format_ch_vec’ ‘format_selectors’
  All user-level objects in a package should have documentation entries.
  See chapter ‘Writing R documentation files’ in the ‘Writing R
  Extensions’ manual.

❯ checking S3 generic/method consistency ... NOTE
  Found the following apparent S3 methods exported but not registered:
    update.step
  See section ‘Registering S3 methods’ in the ‘Writing R Extensions’
  manual.

❯ checking Rd \usage sections ... NOTE
  S3 methods shown with full name in documentation object 'update.step':
    ‘update.step’
  
  The \usage entries for S3 methods should use the \method markup and not
  their full name.
  See chapter ‘Writing R documentation files’ in the ‘Writing R
  Extensions’ manual.


topepo:
I think that you need to update from master. I'm going to do a release this week (for the new `tidyr`). I'm probably going to submit Wednesday.

EdwinTh:
Indeed that resolved it. Could have worked that out myself, sorry for the trouble. Unfortunately did not make it before Wednesday because I was on holiday.

topepo:
I updated from master in #417 to bring these changes in. Thanks!

PR #377, opened 2019-08-23
boilerplate for new steps and checks

EdwinTh:
Created a little function for creating new steps and checks, internal use. Will setup the boilerplate code, so you don't have to copy paste from an existing step or check anymore. Parts to be filled in are marked with `< comment >`


topepo:
Can you add this under `recipes/inst/templates`?

EdwinTh:
Done

PR #381, opened 2019-09-03
added step_xgb_binning which performs discretisation in a supervised way using xgboost

konradsemsch:
- ``` step_xgb_binning ``` runs three different versions of an xgboost model on top of numerical data and chooses the one with the best performance (using internal validation set) and then converts numerical variables into buckets

To dos:
- [ ] Tests

- [x] Print methods

- [ ] Anything else that I missed?

- [ ] tidy and tunable method

- [x] adding options as an argument to control the xgboost runs


topepo:
I think that this might be better in `embed` (which is where we would put _most_ supervised encoding methods). 

If you are interested in doing that, I can add some comments here about the api and a few other things.

konradsemsch:
Yeah, I'm happy to carry it along and complete it wherever you think it makes most sense. 

I would find some comments and guidelines from your side very helpful and I'll take care of the rest as much as I can :)

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
+#' Discretize numeric variables with XgBoost
+#'
+#' `step_xgb_binning` creates a *specification* of a recipe
```
topepo: Can we call it `step_discretize_tree`? It would keep it more consistent with the existing one and avoid being method-specific in case we want a different underlying model at some later date.

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
...
+#'  into bins in a supervised way using an XgBoost model.
+#'
+#' @param recipe A recipe object. The step will be added to the sequence of
+#'   operations for this recipe.
+#' @param ... One or more selector functions to choose which variables are
+#'   affected by the step. See [selections()] for more details.
+#' @param role Defaults to `"predictor"`.
+#' @param trained A logical to indicate if the quantities for preprocessing
+#'   have been estimated.
+#' @param outcome A character string to indicate the outcome variable to train XbBoost models.
```
topepo: This is currently inconsistent with some `recipes` steps, can you use tidy select here? See tidymodels/embed#23

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
...
+#' @param rules The splitting rules of the best XgBoost tree to retain for each variable.
```
topepo: It would be good to have the tuning parameters exposed (instead of hard-coded in `run_xgboost`) as main arguments to the step so make it easy to tune. This will add a lot of people will want to tune it.

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
...
+    data = as.matrix(train[, variable]),
+    label = ifelse(train[[target]] == levels[[1]], 0, 1)
+  )
+
+  xgb_test <- xgboost::xgb.DMatrix(
+    data = as.matrix(test[, variable]),
+    label = ifelse(test[[target]] == levels[[1]], 0, 1)
+  )
+
+  xgb_runs <- purrr::map(c(10, 20, 50), function(x) run_xgboost(xgb_train, xgb_test, x))
```
topepo: Instead of hard-coding these, I would set it up for one value and we can tune it as needed. This will simplify the code a lot too. Please keep the argument names consistent with `parsnip:::boost_tree()`

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
...
+
+  xgb_selected <- if (xgb_best == "mdl_max_bin_50") {
+    xgb_runs[[3]]
+  } else if (xgb_best == "mdl_max_bin_20") {
+    xgb_runs[[2]]
+  } else {
+    xgb_runs[[1]]
+  }
+
+  xgboost::xgb.model.dt.tree(
```
topepo: Maybe return a named list of vectors of bin values (this looks like it is returning the tree object if I understand it).

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
...
+                                  ...) {
+
+  vars <- object$rules
+
+  for(i in seq_along(vars)){
+
+    var <- names(vars)[[i]]
+    binned_data <- new_data
+
+    binned_data[, var] <- cut(
```
topepo: This might make it more complex but think about maybe adding an additional level for missing values.

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
...
+      skip = skip,
+      id = id
+    )
+  }
+
+run_xgboost <- function(.train, .test, .max_bin){
+  xgboost::xgb.train(
+    params = list(
+      eta = 0.3,
+      max_bin = .max_bin,
```
topepo: Please add this as a main argument called `num_breaks`. We can document it as the maximum number of breaks.

R/xgb_binning.R
topepo: Also, you should control the seed. I'd suggest wrapping the call to `run_xgboost` inside of `with_seed`: `withr:::with_seed(sample.int(10%6, 1), run_xgboost())`.

Also also, if you have an `options` argument, please default `nthread = 1`.

R/xgb_binning.R
```diff
@@ -0,0 +1,250 @@
...
+    ),
+    nrounds = 100,
+    data = .train,
+    watchlist = list(
+      train = .train,
+      test  = .test
+    ),
+    tree_method = "hist",
+    early_stopping_rounds = 10,
+    objective = "binary:logistic",
```
topepo: This could conceivably be used for regression models too (right?) so maybe set this based on the class of the original outcome vector.

R/xgb_binning.R
topepo: As noted below, this could also work for regression problems so maybe consider relaxing this a bit.

konradsemsch:
Coolio, let me circle back with questions/ solutions on these in the next 3 days and we'll take it from there!

R/xgb_binning.R
konradsemsch: Makes sense - changing

R/xgb_binning.R
konradsemsch: Then the following ones would be my set of candidates where we pass some reasonable defaults for this step:

eta (0.3)
max_bin (10, 20, 50) <- that's pretty much the most important parameter for this step
max_depth (1)
mind_child_weight (5)
colsample_bytree (0.3)

What do you think about the others such as nround and early_stopping rounds?

R/xgb_binning.R
konradsemsch: That's definitely a good point and I would propose something like this since xgboost cares only about the number of levels of the outcome and actually doesn't run on factor outcomes:

```
# Checking the number of levels of the outcome
  levels <- sort(unique(df[[outcome]]))

  if(any(is.na((levels)))){
    stop("Outcome variable contains missing values.", call. = FALSE)
  }

  # Defining the objective function
  objective <- if (length(levels) == 2){
    "binary:logistic"
  } else if (class(df[[outcome]]) %in% c("double", "integer")){
    "reg:squarederror"
  } else {
    stop("Outcome variable needs to have two levels (binary classification) 
         or be a double/ integer (regression)", call. = FALSE)
  }
```

R/xgb_binning.R
konradsemsch: I just saw your comment below on ```num_breaks``` - so would you expose more of them or only this main one that's critical to this step?

R/xgb_binning.R
konradsemsch: Makes sense - done

R/xgb_binning.R
konradsemsch: Done, although let's align on your comment above on exposing other arguments

R/xgb_binning.R
konradsemsch: I actually like the returned format of this because it allows to clearly see the entire splitting path:

```
> xgboost::xgb.model.dt.tree(
+     model = xgb_mdl,
+     trees = xgb_mdl$best_iteration,
+     use_int_id = TRUE
+   )
   Tree Node Feature Split Yes No Missing      Quality     Cover
1:   11    0       0    19   1  2       1  5.491004470 505.31586
2:   11    1    Leaf    NA  NA NA      NA -0.001171782 467.08981
3:   11    2    Leaf    NA  NA NA      NA  0.115687110  38.22604
```
How about leaving it as is?

R/xgb_binning.R
konradsemsch: After the changes I made this part looks like this now:

```
xgb_mdl <- withr:::with_seed(sample.int(10^6, 1), run_xgboost(xgb_train, xgb_test, num_breaks, objective))
```

What would you change here specifically in the context of ```parsnip:::boost_tree()```?

R/xgb_binning.R
konradsemsch: Hmmm, I'm not sure whether this really belongs here. I'd be probably more in favour of leaving it as is and using ```step_* imputation``` methods before that. Specifically I would also like to work on an imputation step that imputes missing values with a constant e.g.: 99999 which is an approach typically used with tree-based methods. By doing so we can ensure that each step focuses on what it should do in isolation. What do you think?

R/xgb_binning.R
konradsemsch: Ok, so you mean to define the outcome in a similar way as it's currently in this example:

```
embed::step_lencode_bayes(pregnant_fct, outcome = vars(diabetes))
```

topepo: Yes.

R/xgb_binning.R
topepo: I would err on the side of too many arguments. The tuning infrastructure coming up works best when the parameters that affect performance are main parameters. Also, Io would use `learn_rate` instead of `eta`; we try to avoid jargon.

R/xgb_binning.R
topepo: I'm not sure; I would treat `num_breaks` as if it were any other parameter. Maybe I need to kick this around a bit to make a good assessment.

R/xgb_binning.R
topepo: It seems a lot to carry around though and adds complexity. Do you need all of the columns? Can you convert it to a tibble (since we keep everything in `recipes` as tibbles)?

R/xgb_binning.R
topepo: > Specifically I would also like to work on an imputation step that imputes missing values with a constant e.g.: 99999 which is an approach typically used with tree-based methods. 

Oh God no. I've been trying to eradicate this approach since the 80's 😀 Please don't ever do this. 

We have a separate step that can do what I proposed so no sweat.

R/xgb_binning.R
konradsemsch: I think we could stick with Node, Feature, Split, Yes, No & Missing. I would keep these 6 columns for now especially for debugging purposes. When we settle down with this a little bit we could probably further limit them.

R/xgb_binning.R
konradsemsch: > Oh God no. I've been trying to eradicate this approach since the 80's 😀 Please don't ever do this.

Could you elaborate on why you consider this a bad approach?

> We have a separate step that can do what I proposed so no sweat.

What do you have in mind specifically?

R/xgb_binning.R
konradsemsch: > Ok, done.

R/xgb_binning.R
konradsemsch: Point taken on ```learn_rate```. Also for the start, let's include 3 most critical parameters for this task which would be: ```max_bin```, ```eta``` and ```max_depth``` (with appropriately changed naming).

R/xgb_binning.R
konradsemsch: Implemented in line with the comment about parameters I made above.

konradsemsch:
Hi @topepo! I just came back from extended holidays that why you haven't heard from me for a while :) I worked on all the points you mentioned and just pushed another working version. Take a look and let me know what you think and then we'll take it further.

konradsemsch:
Hi @topepo - did you have any time to look through the latest changes? I could probably get going the next points once we have agreement over here.

R/xgb_binning.R
topepo: > Could you elaborate on why you consider this a bad approach?

Special codes, especially ones that try to emulate data of another type, is a poor solution to the problem. 

> What do you have in mind specifically?

Folks can use `step_unknown()` beforehand,

topepo:
Go ahead and move the PR to `embed`. I might make some small changes at that point.

konradsemsch:
Moved this PR to embed: https://github.com/cran/embed/pull/1

PR #409, opened 2019-11-14
Fix typo: sings -> signs

ewv88:
Small typo fix in the `selections.Rd` file

### DIFF
--- a/R/selections.R
+++ b/R/selections.R
@@ -28,7 +28,7 @@
 #'   \item These arguments are not evaluated until the `prep`
 #'    function for the step is executed.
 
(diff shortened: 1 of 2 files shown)


topepo:
Thanks but can you update the sources instead of the Rd file? 

>% Generated by roxygen2: do not edit by hand
> % Please edit documentation in R/selections.R

ewv88:
I reverted the commit on the Rd file and moved it to the source file. Thanks! 
Let me know if this is sufficient.

topepo:
Thanks! Good job!

Issue #157, opened 2018-06-10
Add `step_cut` or equivalent


aaron-horowitz:
`discretize` as it's currently implemented, isn't quite equivalent to base R's `cut` function, in that in the latter you can specify explicit cutpoints _or_ a single integer reflecting the number of intervals.  For interpretability of variables, it's often useful to set boundaries that are more "human readable". Would it be worth it to add a `step_cut` discretization alternative and/or build that capability into the core function itself, a la `cut`.

Since recipes are often meant to be used with multiple variables, this may not ideal, so an alternative is some sort of rounding parameters for cleaner breaks (ie round to neareast N).

EdwinTh:
I think this is a good idea (I need it myself now and then :)). @topepo I am happy to do this if you want to have it added.

juliasilge:
Closed in #452

PR #452, opened 2020-01-16
Adding step_cut

EdwinTh:
Took the liberty to create `step_cut` as discussed  #157, because I really needed it myself. Let me know what you think.


topepo:
Can you look the code over and change some parts to be consistent with the tidyverse style guide? Fir example, the `if` statements in `create_full_breaks` should look like:

```r
create_full_breaks <- function(var, breaks) {
  stopifnot(is.numeric(var), is.numeric(breaks))
  if (min(var) < min(breaks)) {
    breaks <- c(min(var), breaks)
  }
  if (max(var) > max(breaks)) {
    breaks <- c(max(var), breaks)
  }
  sort(breaks)
}
```

Also, please add a `tunable()` method. You might map `breaks` to `dials:::num_breaks()`.

EdwinTh:
Will look into it soon.

EdwinTh:
Just looked into `dial` and `tunable`. I am not sure it is useful to map to `dials:::num_breaks()`, since this `step` is aimed at creating cuts that are meaningful instead of optimizing the number of cuts. 

What I can imagine is that you provide all the possible cuts and that you optimize over which cuts are needed and which not. For example, if you want the cut points only at every decade you provide a grid of, say, six decades and you optimize which decades can be merged. Would that something you can use `dial` and `tunable` with.

topepo:
> Just looked into `dial` and `tunable`. I am not sure it is useful to map to `dials:::num_breaks()`, since this `step` is aimed at creating cuts that are meaningful instead of optimizing the number of cuts. 

Good point. Let's forgo that then.

topepo:
Sorry for the delay on this. Can you update form master?

topepo:
I'll merge. I need to do a release today or tomorrow.

EdwinTh:
Sorry, was a little under the weather last couple of days, could not look into it. Thanks

PR #455, opened 2020-01-30
Fixing the broken link to "The R Formula Method: The Good Parts"

doug-friedman:


DavisVaughan:
Can you please change the Rmd file, and re render the md file using knitr? We do not alter the md files directly

doug-friedman:
🤦‍♂ Sorry about that! Will do.

doug-friedman:
I think I'm all set?

DavisVaughan:
Thank you!

Issue #443, opened 2020-01-09
rename yj_trans to avoid conflicts


topepo:
The `scales` package also has a function called `yj_trans`. They both export is so there is often the report:

```
## x recipes::yj_trans() masks scales::yj_trans()
```

We can just rename ` recipes::yj_trans` to ` recipes::yj_transform`

PR #457, opened 2020-01-31
Raname yj transform

RaymondBalise:
Addressed issue #443 renamed yr_trans to yr_transform


DavisVaughan:
Since this is a breaking change, can you please add a news bullet describing that `yj_trans()` has been renamed?

PR #460, opened 2020-01-31
Add ID interface to `tidy.recipe()`, with corresponding tests.

### LINKED ISSUE #273: Add ID interface to `tidy()` method for steps
At rstudioconf, I got a question about extracting the information from an individual estimated step. For example, with box cox, maybe you wanted to know the exact value of lambda for the estimated step.

We could have a new verb that accepts a prepped recipe, and extracts the step information by `id` name. It would be as simple as `rec$steps[[1]]` for the first step (or a more specialized way that finds the right id name and returns that step).

I don't think we need to return the information in any special way, just return the list of info corresponding to that step, and let the user process it however they want.

smingerson:
Add an `id` argument to `tidy.recipe()` for selecting a step. Only allows once step, which matches how the `number` argument works. Only `number` or `id` may be provided, not both. 

Fixes #273

### DIFF
--- a/R/tidy.R
+++ b/R/tidy.R
@@ -18,7 +18,7 @@
 #'  `operation` (either "step" or "check"),
 #'  `type` (the method, e.g. "nzv", "center"), a logical
 #'  column called `trained` for whether the operation has been
-#'  estimated using `prep`, a logical for `skip`, and a character column `id`. 
+#'  estimated using `prep`, a logical for `skip`, and a character column `id`.
 #'
 #' @examples
 #' library(modeldata)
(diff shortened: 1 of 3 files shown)


DavisVaughan:
You will need to document the `id` parameter somewhere around here https://github.com/tidymodels/recipes/blob/50cbcbeb18dddaf1c6d041d605363fb293da8509/R/tidy.R#L10

R/tidy.R
```diff
@@ -44,11 +44,27 @@ NULL
...
+  # add id = NA as default. If both ID & number are non-NA, error.
+  # If number is NA and ID is not, select the step with the corresponding
+  # ID. Only a single ID is allowed, as this follows the convention for number
   num_oper <- length(x$steps)
   if (num_oper == 0)
     stop("No steps in recipe.", call. = FALSE)
   pattern <- "(^step_)|(^check_)"
+  if (!is.na(id)) {
+    if (!is.na(number))
+      rlang::abort(paste0("You must specify one of `number` or `id`, ",
```
DavisVaughan: Maybe: "You may specify `number` or `id`, but not both."

R/tidy.R
```diff
@@ -44,11 +44,27 @@ NULL
...
+                          "or neither."))
+    if (is.null(id)) {
```
DavisVaughan: Probably can be more robust with `if (length(id) != 1L && !is.character(id))`

R/tidy.R
```diff
@@ -44,11 +44,27 @@ NULL
...
+      rlang::abort("If `id` is provided, it must be a length 1 character vector.")
+    }
+    step_ids <- vapply(x$steps, function(x) x$id, character(1))
+    if(!(id %in% step_ids)) {
+      rlang::abort("`id` provided not found in the recipe.")
```
DavisVaughan: Maybe flip this around, "Supplied `id` not found in the recipe"

smingerson:
I've added documentation for `id` and updated that for `number` and the return value.

DavisVaughan:
This looks good to me, but ill let @topepo give the final review here

Issue #282, opened 2019-01-29
Suggestion: add a keep_vars arguments (for some steps)


ericlecoutre:
I am thinking in particular on multivariate steps such as pca or ica.
Sometimes, I am interested in feature engineering so that I would like to generate PCA plus keep source variables.
As PCA allows adding prefix such as "PCA_", one may still manipulate source variables with a `-starts_with("PCA")`.
Same for `embed` package.

topepo:
We'll add a `retain` options for 

- [ ] `step_pca`
- [ ] `step_ica`
- [ ] `step_nnmf`
- [ ] `step_kpca`
- [ ] `step_pls`
- [ ] `step_spatialsign`
- [ ] `step_isomap`

This should be pretty straightforward

jonthegeek:
Working on this for tidy-dev-day

PR #462, opened 2020-01-31
Adding keep_vars option to various steps

jonthegeek:
This is still in progress, but I want to get some feedback before I implement this 6x more times.


jonthegeek:
Closes #282.

R/pca.R
```diff
@@ -155,6 +159,8 @@ step_pca_new <-
 
 #' @export
 prep.step_pca <- function(x, training, info = NULL, ...) {
+  x <- add_new_args(x)
```
DavisVaughan: I think the way this works is a bit too complicated.

Maybe we could just use a check for `is.null(x$keep_vars)` in the `prep()` and `bake()` methods for these steps, and set it to `FALSE` if it is NULL, but then throw a warning that the user should regenerate their recipe because the arguments have been updated?

jonthegeek: Yeah, that's what I did at first (with %||%), but I had the beginnings of a more reusable idea walking back from dev day and wanted to give it a shot. I agree it's probably a bit too fragile right now, but wanted to log the general idea to get gears turning for a prep/bake backward compatibility safety valve. I'll go back to the other method.

jonthegeek: Ok, I rolled back the over-complicated change and added a warning. Can you review the `step_pca` implementation to make sure everything looks in-order? If so, I'll apply it to the other 6 functions where it makes sense.

dgkf:
I just started using `recipes` and quickly ran into issues not being able to retain the source variables through steps, requiring a number of hoops in the form of `step_mutate_at` to duplicate a set of variables before processing them.

This looks like it's addressing exactly that problem, but there hasn't been much recent activity. Is this still being actively developed?

juliasilge:
@jonthegeek thank you for your patience on this, which I think was from tidyverse dev day at the beginning of the year?! 😳 I know you were waiting on a check from one of us but are you still interested in finishing this PR for `keep_vars`?

jonthegeek:
> @jonthegeek thank you for your patience on this, which I think was from tidyverse dev day at the beginning of the year?! 😳 I know you were waiting on a check from one of us but are you still interested in finishing this PR for `keep_vars`?

I'd be happy to work on it again, but the review would definitely be helpful, and then I'll need a minute to remember what the heck I was doing 😋

topepo:
It looks good to me. Some minor things: 

Can you use single quotes around arguments and back-ticks around function names in messages? For example: 

```r
 "'keep_vars' was added to `step_pca()` after this recipe was created."
```

Also, let's use the argument name `keep_original_cols` instead. A little more descriptive.

It would be good to write a reusable function for the code block: 

```r
check_keep_original_cols <- function(object) {
   if (is.null(object$keep_original_cols)) {
      object$keep_original_cols <- FALSE
      rlang::warn(
         paste(
            "'keep_original_cols' was added to `step_pca()` after this recipe was created.",
            "Regenerate your recipe to avoid this warning.",
            sep = "\n"
         )
      )
   }
   object
}
```

then use

```r
object <- check_keep_original_cols(object)
```

or 

```r
x <- check_keep_vars(x)
```

as appropriate. 

Can you write a test for old recipes too? You could create one then set it to NULL for testing.

juliasilge:
@jonthegeek are you interested in finishing up this PR based on the above feedback? We would like to get this feature merged in to recipes and could possibly pick it up ourselves, if there are blockers to you finishing it up. It looks great so far! 🙌

jonthegeek:
@juliasilge Soooo. I have time constraints... but I also seem to have deleted this repo in a purge of (what I believed were) "unused" repos, so I need to go back in and sort out the code. So if you want to just take what I started (as shown in this changelog) and go from there, that's fine, because I keep not finding time to work on this.

juliasilge:
Ah wow, yes, I can't even pull this PR down at all myself to update it. I do think we'll have to close this PR and open a new one based on your start here.

Thank you so much for your contribution! 🙌

jonthegeek:
Yeah, sorry that I made a mess of things! I'm glad you were able to clean it up!

PR #461, opened 2020-01-31
[WIP] Better naming step funcs

gdequeiroz:
Related to #262 
I started working on this a year ago and never finished it.  @labouz will pick this up.

Issue #262, opened 2018-11-23
Better naming for step_classdist and step_depth


topepo:
From #261 add an options similar to `step_dummy` or `step_pca` that allows for more flexibility when naming the new variables. Do the same for `step_depth`.

gdequeiroz:
On it!

gdequeiroz:
Progress
--------------------------------
- [ ] Add argument prefix to the`step_classdist` function
- [ ] Add document to reflect the change
- [ ] Add tests
- [ ] Add argument prefix to the `step_depth` function
- [ ] Add document to reflect the change
- [ ] Add tests

juliasilge:
This is closed in #468. Thank you so much @gdequeiroz! 👯‍♀️

gdequeiroz:
@juliasilge Thank you for doing the work and closing this!

PR #468, opened 2020-02-10
More flexible column names for step_classdist and step_depth

juliasilge:
This PR addresses #262 and builds on the excellent work of @gdequeiroz ✨ in #461.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -14,3 +14,4 @@ tests/testthat/NMF*
 ^azure-pipelines\.yml$
 ^pkgdown$
 ^\.github/workflows/R-CMD-check\.yaml$
+^\.github/workflows/pr-commands\.yaml$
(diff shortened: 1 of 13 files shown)


juliasilge:
After I updated the GH actions, the checks were successful **except** that `step_poly()` now is breaking on R 3.2.

juliasilge:
OMG the GH actions all work but now Travis is broken from some LaTeX issue?? `Overfull \hbox` everywhere...

juliasilge:
I thought this PR might be haunted but it seems ready to go now.

juliasilge:
I mentioned this in a commit message, but so you see it here as well:

I added the `stringsAsFactors` change for recipes to this PR in 3974ff0bbc732ada7c29755ed8bb37682bae2586 so that it would pass the checks. In hindsight, maybe it would have been better to put that in a separate PR, since it's a separate thing.

PR #478, opened 2020-03-04
switch model to step in recipes_pkg_check

EmilHvitfeldt:


topepo:
I think that the 3.6 failure is an artifact. The other is related to `stringsAsFactors()` change. I'll merge. 

Thanks

PR #490, opened 2020-04-03
changes for new tibble release and dplyr 1.0.0

juliasilge:
I'm starting a PR with changes needed for both the new (already out) tibble release and dplyr 1.0.0. Currently WIP and addresses  #479.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -56,5 +56,5 @@ License: GPL-2
 VignetteBuilder: knitr
 Encoding: UTF-8
 LazyData: true
-RoxygenNote: 7.0.2.9000
+RoxygenNote: 7.1.0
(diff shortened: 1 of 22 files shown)


juliasilge:
Some checks are failing because xml2, dplyr, tidyr, etc are not available on those infrastructures. Other than that, this is ready to go! 🚀 

Most of these changes are due to 2️⃣ updates:

- the [inner names](https://github.com/tidyverse/tibble/blob/master/NEWS.md#major-breaking-changes) that are now allowed in tibble (this is already on CRAN as of today)
- the [removal of `dplyr:::all.equal.tbl_df`](https://github.com/tidyverse/dplyr/blob/master/NEWS.md#breaking-changes), which ignored column order (not yet on CRAN as of today)

Issue #489, opened 2020-03-17
missing values in predictions even with step_naomit()


topepo:
From [this post](https://community.rstudio.com/t/recipe-and-step-naomit-error-with-grid-tune/55540): 

``` r
library(tidyverse)
library(tidymodels)
#> ── Attaching packages ──────────────────────────────────────────────────────────────────────────────────── tidymodels 0.1.0 ──
#> ✓ broom     0.5.4          ✓ rsample   0.0.5     
#> ✓ dials     0.0.4          ✓ tune      0.0.1.9000
#> ✓ infer     0.5.1          ✓ workflows 0.1.0     
#> ✓ parsnip   0.0.5          ✓ yardstick 0.0.5     
#> ✓ recipes   0.1.9.9000
#> ── Conflicts ─────────────────────────────────────────────────────────────────────────────────────── tidymodels_conflicts() ──
#> x scales::discard() masks purrr::discard()
#> x dplyr::filter()   masks stats::filter()
#> x recipes::fixed()  masks stringr::fixed()
#> x dplyr::lag()      masks stats::lag()
#> x dials::margin()   masks ggplot2::margin()
#> x yardstick::spec() masks readr::spec()
#> x recipes::step()   masks stats::step()

set.seed(1234)
mtcars_tb <- mtcars %>%
  as_tibble() %>%
  mutate(vs = c(sample(vs, 22), rep(NA_integer_, 10)))

set.seed(1234)
cv_fold_mtc <- vfold_cv(mtcars_tb)

lasso_mod <-
  linear_reg() %>%
  set_engine("glmnet") %>%
  set_args(penalty = tune(),
           mixture = 1)

rec <- recipe(mpg ~ disp + vs,
              data = mtcars_tb) %>%
  step_naomit(everything()) %>%
  step_dummy(all_nominal()) %>%
  step_normalize(all_numeric())

tune_grid(
  lasso_mod,
  rec,
  resamples = cv_fold_mtc,
  control = tune::control_resamples(verbose = TRUE,
                                    save_pred = TRUE)
)
#> i Fold01: recipe
#> ✓ Fold01: recipe
#> i Fold01: model 1/1
#> ✓ Fold01: model 1/1
#> i Fold01: model 1/1 (predictions)
#> x Fold01: model 1/1 (predictions): Error: Column `.row` must be length 2 (the numb...
#> i Fold02: recipe
#> ✓ Fold02: recipe
#> i Fold02: model 1/1
[...]

PR #512, opened 2020-05-17
changed skip defaults for 4 steps

topepo:
Changes for tidymodels/tune#181 and tidymodels/tune#206

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,10 @@
+# recipes 0.1.11.9000
+
+## Breaking Changes
+
+* `step_filter()`, `step_slice()`, `step_sample()`, and `step_naomit()` had their defaults for `skip` changed to `TRUE`. In the vast majority of applications, these steps should not be applied to the test or assessment sets. 
+
+
 # recipes 0.1.11
 
 ## Other Changes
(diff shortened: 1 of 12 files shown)


topepo:
closes #489

closes tidymodels/tune#181 

closes tidymodels/tune#206

dkrozelle:
Hi @topepo, sorry if this is obvious, I'm a bit new to using recipes. I'm using tune_grid with a recipe that includes step_naomit(). With skip=TRUE we retain NA in each bootstrap (linear_reg doesn't throw an error but I'm suspicious?), with skip=FALSE they remain (and tune_grid throws an error due to different bootstrap sample sizes). Would it be preferable to preprocess to remove NA rows before creating bootstrap splits or is there another method? Thanks!

juliasilge:
Hello there @dkrozelle! In the future, it will probably be better if you open a new issue or post a question on [RStudio Community](https://rstd.io/tidymodels-community); commenting on closed issues or merged PRs isn't the best for getting help.

Removing `NA` values when you are going to use resampling is pretty tricky. You can check out the stronger advice we are thinking about adding [in this PR](https://github.com/tidymodels/recipes/pull/590):

> In most instances that affect the rows of the data being predicted, this step probably should not be applied at all

The best use of recipe steps are when you need to estimate some preprocessing steps from training data and then apply them to new data like the assessment sets of resamples, testing data, new data at prediction time, etc.

PR #559, opened 2020-08-24
Update step_pca help documentation

irkaal:
A small fix to the step_pca help documentation to be consistent with the actual error.


topepo:
Thanks!

PR #560, opened 2020-08-25
Small fix for step_ns help documentation

irkaal:
Update title for step_ns help documentation to match the title in ns.R


juliasilge:
Related to #557 

Thanks so much @irkaal!

PR #590, opened 2020-10-21
More opinionated advice on row sampling steps

juliasilge:
Related to tidymodels/tune#241

The row filtering steps are still tripping some folks up so let's make the documentation a bit stronger and clearer in what situations they are a problem. Take a look and see what feedback you have on this little blurb I made a roxygen template for, to go in any row filtering step. I added it to:

- `step_filter()`
- `step_naomit()`
- `step_sample()`
- `step_slice()`

This, combined, with the new error message in tidymodels/tune#307, hopefully will provide some significant help.

### DIFF
--- a/R/filter.R
+++ b/R/filter.R
@@ -3,6 +3,7 @@
 #' `step_filter` creates a *specification* of a recipe step
 #'  that will remove rows using [dplyr::filter()].
 #'
+#' @template row-ops
 #' @inheritParams step_center
 #' @param ... Logical predicates defined in terms of the variables
 #'  in the data. Multiple conditions are combined with `&`. Only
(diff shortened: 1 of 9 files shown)


DavisVaughan:
Should this also be in `themis::step_downsample()`?

juliasilge:
I don't know; the use case seems pretty different and I don't know that we'd want to say something like:

> In most instances that affect the rows of the data being predicted, this step probably should not be applied at all

about the subsampling steps. 

The way that the subsampling steps are implemented don't seem to cause folks trouble FWIW (the default is `skip = TRUE`, people don't use them for anything weird beyond their intended purpose) while these row filtering steps here _are_ causing problems 👻 to crop up for users.

PR #593, opened 2020-10-28
Minor consistancy changes

wfjvdham:
- Other steps are exported so I think center should be exported as well.
- It is more difficult to find the generic with these extra spaces so I removed them.

Also I think it is helpful to explain [here](https://www.tidymodels.org/learn/develop/recipes/) that you need to import prep() and bake() if you make your own step function


juliasilge:
Thanks so much @wfjvdham! I added a note that you can [preview at the bottom of this section](https://deploy-preview-201--tidymodels-org.netlify.app/learn/develop/recipes/#create-the-bake-method). Do you think that would be helpful to folks?

wfjvdham:
Yes thanks! Makes it more clear for me.

juliasilge (approved):

PR #594, opened 2020-10-30
Reexport tunable + required_pkgs from generics

### LINKED ISSUE #592: required_pkgs not working for recipe steps
The function `tune::required_pkgs()` is not finding the S3 methods for individual recipe steps. It seems like it is finding `required_pkgs.recipe` and `required_pkgs.step` ✅  (both of which are in the tune package) but not methods for individual steps 🚫  (which are in the recipes package).

``` r
library(tidymodels)

pls_recipe <-
  recipe(mpg ~ ., data = mtcars) %>%
  step_pls(all_predictors(), outcome = "mpg")

required_pkgs(pls_recipe, FALSE)
#> character(0)

kpca_rec <- recipe(mpg ~ ., data = mtcars) %>%
  step_YeoJohnson(all_predictors()) %>%
  step_normalize(all_predictors()) %>%
  step_kpca_rbf(all_predictors())

required_pkgs(kpca_rec, FALSE)
#> character(0)
```

<sup>Created on 2020-10-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

Related to tidymodels/extratests#10

juliasilge:
This PR (using the development version of generics) solves our S3 struggles by reexporting `tunable()` and `required_pkgs()` from generics. This closes #592:

``` r
library(tidymodels)
#>   method       from   
#>   tunable.step recipes

pls_recipe <-
  recipe(mpg ~ ., data = mtcars) %>%
  step_pls(all_predictors(), outcome = "mpg")

required_pkgs(pls_recipe, FALSE)
#> [1] "mixOmics"

kpca_rec <- recipe(mpg ~ ., data = mtcars) %>%
  step_YeoJohnson(all_predictors()) %>%
  step_normalize(all_predictors()) %>%
  step_kpca_rbf(all_predictors())

required_pkgs(kpca_rec, FALSE)
#> [1] "dimRed"  "kernlab"
```

<sup>Created on 2020-10-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

A question ❓ still remaining: there is a method for `tunable.step` in _both_ tune and recipes. Is that right?


DavisVaughan:
>  there is a method for tunable.step in both tune and recipes.

That smells wrong

DESCRIPTION
```diff
@@ -59,3 +59,5 @@ LazyData: true
 RoxygenNote: 7.1.1.9000
 Roxygen: list(markdown = TRUE)
 RdMacros: lifecycle
+Remotes:
+    r-lib/generics
```
DavisVaughan, line 63: Reminder: 0.1.0 version should be on cran now

R/zzz.R
```diff
@@ -22,7 +22,7 @@ maybe_register_S3_methods <- function() {
 
   for (i in seq_along(tunable_names)) {
     class <- tunable_classes[[i]]
-    s3_register("tune::tunable", class)
+    s3_register("generics::tunable", class)
   }
```
DavisVaughan, line 26: Now that you directly register the S3 methods with the `@export` tag, I don't think we need to lazy register them now. You probably don't need `tunable_names` from above either

R/zzz.R
```diff
@@ -58,7 +58,7 @@ maybe_register_S3_methods <- function() {
 
     for (i in seq_along(req_pkgs_names)) {
       class <- req_pkgs_classes[[i]]
-      s3_register("tune::required_pkgs", class)
+      s3_register("generics::required_pkgs", class)
     }
   }
```
DavisVaughan, line 63: Same comment as before, we probably don't need to lazy register them now. We shouldn't need the if statement requiring a minimum tune version either. The whole chunk can probably be removed

DavisVaughan (changes requested):

DESCRIPTION
```diff
@@ -59,5 +59,3 @@ LazyData: true
 RoxygenNote: 7.1.1.9000
 Roxygen: list(markdown = TRUE)
 RdMacros: lifecycle
-Remotes:
```
DavisVaughan: You may want to pin to the new version `generics (>= 0.1.0)`

juliasilge: ✅

R/zzz.R
juliasilge: Thanks for this explanation; I am clearer now on how this registering is working.

R/zzz.R
```diff
@@ -6,27 +6,14 @@
 }
 
 # This package has specific methods for the `tunable()` and `required_pkgs()`
-# generic. That generic is defined in the `tune` package. As of R 4.0, we need to register them.
+# generic. That generic is defined in the `generics` package. As of R 4.0, we need to register them.
```
DavisVaughan, line 9: Can remove outdated comments

DavisVaughan (approved):

DavisVaughan:
LGTM now!

R/zzz.R
juliasilge: A question for you... is there something different about the `tidy` methods for steps and checks that we _do_ need to lazily register them, even though we are reexporting `tidy` from generics also? For example, see lines 15-24 or so.

DavisVaughan: Oh interesting. Probably not. You can probably remove the lazy registration of those too. They may have been from before we had generics, and had to Suggest broom and lazy register methods there

Issue #567, opened 2020-09-14
Recipes eats too much RAM


LordRudolf:
Hello,

Thank you for the great package!

The "recipes" eases and automates the data preprocessing in a very high degree. Unfortunately, it comes with the costs of high RAM usage. 

In the example bellow, manually preprocessed data and model built via "caret" takes up to 201 MB whereas doing the same procedures via "recipes" takes up to 665 MB. In practice, where the work is with much bigger datasets (with the dimensions 50 000 x 500) and many more preprocessing steps, the difference is not 3-fold but already a 10-fold (solely "caret" model takes 500 MB of RAM whereas the equivalent "recipes" model takes about 5 GB). Unfortunately, the following RAM issue prevents us to use the "recipes" in the production environment where we host tens of different model APIs. 

I will be grateful to hear the solutions for the RAM issue.

```
library(carData)
library(caret)
library(forcats)

data(MplsStops)
df <- MplsStops

##Preprocesing
for(i in 1:ncol(df)) {
  if(is.factor(df[, i])) {
    levels(MplsStops[,i]) <- make.names(levels(MplsStops[,i]))
    df[, i] <- fct_explicit_na(MplsStops[, i], na_level = "NA_value")
  }
}

##Train caret model
the_control <- trainControl(
  method = 'none',
  summaryFunction = twoClassSummary,
  classProbs = TRUE,
  savePredictions = FALSE
)

fit <- train(problem ~ .,
             data = df[, !colnames(df) %in% c('idNum', 'date')],
             method = 'ranger',
             metric = 'ROC',
             tuneGrid = data.frame(mtry = 4, splitrule = 'gini', min.node.size = 1),
             trControl = the_control
             )

m1 <- pryr::mem_used()
m1 <- pryr::mem_used()


##Preprocessing and training with recipes
rm(fit, df)
.rs.restartR() #Restart the R session now
df <- MplsStops

library(caret)
library(recipes)
rec <- recipe(problem ~ ., data = df) %>%
  update_role(idNum, date, new_role = 'non_predictor') %>%
  step_unknown(all_nominal(), -all_outcomes(), new_level = 'NA_value')

fit <- train(rec,
             data = df,
             method = 'ranger',
             metric = 'ROC',
             tuneGrid = data.frame(mtry = 4, splitrule = 'gini', min.node.size = 1),
             trControl = the_control
)

m2 <- pryr::mem_used()
```

juliasilge:
Related to #548 

[...]

Issue #463, opened 2020-02-01
Explain deg_free argument in step_ns()


EmilHvitfeldt:
A quick intuition-based explanation of deg_free would be nice to have in the documentation.

juliasilge:
Closed in #613

PR #613, opened 2020-12-04
Update documentation for many variable situations, step_interact()

### LINKED ISSUE #548: C-stack achieved at quickly as the number of variables increase
I started looking into the recipes package, which seems to provide some incredible tools for not only preprocessing but also evaluating at a production level, if speed is not of the essence.  Especially the latter part is something that is extremely useful. 

In my current line of work however I often work around 300 variables, and this quickly turns out to be a problem for the recipes, as it hits the stack limit rather quickly

```r
library(recipes)
nmax <- 300
data <- matrix(rnorm(nmax), ncol = nmax)
data <- as.data.frame(data)
#Success vector
v <- logical(nmax - 1)
for(i in seq_len(nmax)[-1]){
  e <- try(recipe(data, reformulate(paste0('V', seq(2, i), collapse = '+'), response = 'V1')), silent = TRUE)
  v[i - 1] <- !inherits(e, 'try-error')
  rm(e)
  gc()
}
#At which point does recipes fail?
which.min(unlist(v))
# [1] 144
```

The problem seems to be how ´recipes:::fun_calls` recursively evaluates formulas.
```r
form <- reformulate(paste0('V', 2:200, collapse = '+'), 'V1')
recipes:::fun_calls(form)
#Error: C stack usage  7976588 is too close to the limit
```

while I am not entirely sure how this could be avoided, it seems like it should be possible to `split` the formula and evaluate in a less recursive way.

This was experienced in a Docker container (using `rocker/tidyverse:4.0.0`), with an allocated memory of 6 GB, 1.5 GB swap memory, 60 GB disk space and 4 CPU's. 

```r
SessionInfo()
R version 4.0.0 (2020-04-24)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04 LTS

Matrix products: default
BLAS/LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/libopenblasp-r0.3.8.so

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C               LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8     LC_MONETARY=en_US.UTF-8    LC_MESSAGES=C             
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                  LC_ADDRESS=C               LC_TELEPHONE=C             LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] recipes_0.1.13 dplyr_1.0.0   

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.4.6         plyr_1.8.6           pillar_1.4.4         compiler_4.0.0       gower_0.2.1          class_7.3-16         iterators_1.0.12     tools_4.0.0         
 [9] rpart_4.1-15         ipred_0.9-9          nlme_3.1-147         gtable_0.3.0         lubridate_1.7.8      lifecycle_0.2.0      tibble_3.0.3         lattice_0.20-41     
[17] pkgconfig_2.0.3      rlang_0.4.7          Matrix_1.2-18        foreach_1.5.0        DBI_1.1.0            rstudioapi_0.11      prodlim_2019.11.13   stringr_1.4.0       
[25] withr_2.2.0          pROC_1.16.2          generics_0.0.2       vctrs_0.3.2          stats4_4.0.0         grid_4.0.0           nnet_7.3-13          caret_6.0-86        
[33] tidyselect_1.1.0     data.table_1.12.8    glue_1.4.1           R6_2.4.1             survival_3.1-12      lava_1.6.7           reshape2_1.4.4       ggplot2_3.3.2       
[41] purrr_0.3.4          magrittr_1.5         ModelMetrics_1.2.2.2 scales_1.1.1         codetools_0.2-16     ellipsis_0.3.0       MASS_7.3-51.5        splines_4.0.0       
[49] colorspace_1.4-1     timeDate_3043.102    stringi_1.4.6        munsell_0.5.0        crayon_1.3.4
```

For completeness sake, this error can be avoided completely by using the `roles` argument rather than a formula., but it seems inconvenient to have one functionality simply crash the function.

### LINKED ISSUE #610: better document step_interact
Mention more about dummy variables and suggest using `starts_with()` when dummy variables have been created. 

We've shown this in a few places but [this question](https://community.rstudio.com/t/recipes-package-cannot-create-interaction-term-in-step-interact/89860) indicates that we can do more. 

Use an example that uses dummy variables instead of the biomass data.

juliasilge:
This PR contains two updates to documentation:

## Advice to avoid formula interface for high-dimensional data

Closes #548 and #567 
The changes in the documentation for `recipe()` guide folks to use `update_role()` instead of the formula interface when they have many variables, to avoid problems with excessive memory use.

## Info on selectors for `step_interact()`

Closes #610
This new example and improvements to documentation more explicitly show folks how to use `starts_with()` when they have dummy variables.

### DIFF
--- a/R/bs.R
+++ b/R/bs.R
@@ -15,7 +15,10 @@
 #'  used as predictors in a model.
 #' @param objects A list of [splines::bs()] objects
 #'  created once the step has been trained.
-#' @param deg_free The degrees of freedom.
+#' @param deg_free The degrees of freedom for the spline. As the
+#'  degrees of freedom for a spline increase, more flexible and
+#'  complex curves can be generated. When a single degree of freedom is used,
+#'  the result is a rescaled version of the original data.
 #' @param degree Degree of polynomial spline (integer).
 #' @param options A list of options for [splines::bs()]
 #'  which should not include `x`, `degree`, or `df`.
--- a/R/interactions.R
+++ b/R/interactions.R
@@ -63,24 +63,24 @@
 
 #' @examples
 #' library(modeldata)
-#' data(okc)
-#' okc <- okc[complete.cases(okc),]
+#' data(penguins)
+#' penguins <- penguins %>% na.omit()
 #'
-#' rec <- recipe(~ diet + age + height, data = okc)
+#' rec <- recipe(flipper_length_mm ~., data = penguins)
 #'
 #' int_mod_1 <- rec %>%
-#'   step_interact(terms = ~ age:height)
+#'   step_interact(terms = ~ bill_depth_mm:bill_length_mm)
 #'
 #' # specify all dummy variables succinctly with `starts_with()`
 #' int_mod_2 <- rec %>%
-#'   step_dummy(diet) %>%
-#'   step_interact(terms = ~ age:starts_with("diet"))
+#'   step_dummy(sex, species, island) %>%
+#'   step_interact(terms = ~ body_mass_g:starts_with("species"))
 #'
-#' int_mod_1 <- prep(int_mod_1, training = okc)
-#' int_mod_2 <- prep(int_mod_2, training = okc)
+#' int_mod_1 <- prep(int_mod_1, training = penguins)
+#' int_mod_2 <- prep(int_mod_2, training = penguins)
 #'
-#' dat_1 <- bake(int_mod_1, okc)
-#' dat_2 <- bake(int_mod_2, okc)
+#' dat_1 <- bake(int_mod_1, penguins)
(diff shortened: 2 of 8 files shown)


juliasilge:
I added one more change to explain `deg_free` in `step_ns()` better, closing #463.

PR #630, opened 2021-01-12
Groupwise recipes

dfalbel:
Would this be interesting to have in recipes? The idea is to be able to add a recipe specification to be applied separately for each group.

For example:

 ```
df <- tibble::tibble(
    g = sample(letters[1:3], size = 100, replace = TRUE),
    x = runif(100),
    y = rnorm(100)
  )

  grouped_rec <- recipes::recipe(y ~ ., data = df) %>%
    recipes::step_center(x, y)

  rec <- recipes::recipe(y ~ ., df) %>%
    step_group(g, grouped_rec)

  rec <- recipes::prep(rec, df)

  out <- bake(rec, df)
```

In my use case I have a dataset in the long-format with multiple time series and I wanted to normalize each series separately because they have very different mean/std. A similar use case appeared on [SO](https://stackoverflow.com/questions/65406436/apply-step-procedure-in-tidymodels-on-a-per-group-basis) recently too.

### CODE REVIEW

**Comment by topepo** [comment]
I can see the need for something like this based on your description of the data/problem. I'm not sure that I want recipes to have this general of a solution though. I think that people would, more often or not, accidentally misuse it and get themselves into trouble (perhaps without realizing it). Reading that SO link, I can't imagine a situation where this is a good idea. 

It would be better to have a `step_center_grouped()` type solution for your specific needs. 

Maybe knowing more about the issue would help too. From reading your description, it might make sense to use a list column or something for the different series.

**Comment by dfalbel** [comment]
Yeah, I can see it might not be a good idea for recipes. 

The use case for me is just normalizing the target variable across a grouping variable when I have a dataset like this:

```
tibble [9,171,228 × 8] (S3: tbl_df/tbl/data.frame)
 $ id              : chr [1:9171228] "MT_001" "MT_001" "MT_001" "MT_001" ...
 $ date            : POSIXct[1:9171228], format: "2012-01-01 00:00:00" ...
 $ pwr_usage       : num [1:9171228] 12.7 16.5 19 17.8 19 ...
 $ hours_from_start: int [1:9171228] 8760 8761 8762 8763 8764 8765 8766 8767 8768 8769 ...
 $ days_from_start : int [1:9171228] 365 365 365 365 365 365 365 365 365 365 ...
 $ hour            : int [1:9171228] 0 1 2 3 4 5 6 7 8 9 ...
 $ day_of_week     : num [1:9171228] 1 1 1 1 1 1 1 1 1 1 ...
 $ month           : num [1:9171228] 1 1 1 1 1 1 1 1 1 1 ...
```

But I can definitely live with nesting the ids, baking and unesting again.

**Comment by topepo** [comment]
So would all of those data series go into the model? Are those outcomes or predictors?

**Comment by dfalbel** [comment]
They will be used as both predictors and outcomes. This dataset will eventually become an array with: `(n_series, n_time_steps, n_cols)`

I will then take running samples of it with something like:

```
predictors <- x[,1:10,]
outcome <- x[,11,"pwr_usage"]
```

topepo:
Got it. Thanks

PR #634, opened 2021-01-15
all_nominal_predictors() and all_numeric_predictors()

### LINKED ISSUE #620: Challenges using `all_nominal()` and `all_numeric()`
We have lots of examples of folks having trouble with `all_nominal()` and `all_numeric()` when they accidentally select outcomes or other non-predictors. Mine talked about running into this during teaching, and here is [one example of many from the general user base](https://stackoverflow.com/questions/65180483/trying-to-predict-the-probability-of-a-binary-variable-being-equal-to-1-using-ti/65221996#65221996). A significant challenge is that because of how/when these steps are evaluated, the error message is often opaque.

What would be a good way to reduce this friction?

- Should `all_nominal()` and `all_numeric()` only select predictors?
- Should these be replaced by new selectors like `all_nominal_preds()` and `all_numeric_preds()`?
- Some recipe steps are more prone to this problem, like `step_dummy()`. Protections at the step level?
- Something else?

topepo:
closes #620


R/selections.R
```diff
@@ -204,6 +204,19 @@ all_predictors <- function() {
   has_role("predictor")
 }
 
+#' @export
+#' @rdname has_role
+all_numeric_predictors <- function() {
```
DavisVaughan, line 209: Add some docs for these two in `?has_role`?

vignettes/Simple_Example.Rmd
```diff
@@ -105,20 +107,20 @@ imputed
 
 It is important to realize that the _specific_ variables have not been declared yet (as shown when the recipe is printed above). In some preprocessing steps, variables will be added or removed from the current list of possible variables. 
 
-Since some predictors are categorical in nature (i.e. nominal), it would make sense to convert these factor predictors into numeric dummy variables (aka indicator variables) using `step_dummy`. To do this, the step selects all predictors then removes those that are numeric:
+Since some predictors are categorical in nature (i.e. nominal), it would make sense to convert these factor predictors into numeric dummy variables (aka indicator variables) using `step_dummy`. To do this, the step selects all non-numeric predictors:
```
DavisVaughan, line 110:
```suggestion
Since some predictors are categorical in nature (i.e. nominal), it would make sense to convert these factor predictors into numeric dummy variables (aka indicator variables) using `step_dummy()`. To do this, the step selects all non-numeric predictors:
```

DavisVaughan (approved):

R/selections.R
topepo: Thats the topic that has all of the selector information.

topepo: I changed the docs to mention them more in ways that the user will see.

juliasilge (approved):
I looked through this and I have no changes to suggest! This is going to solve a lot of problems for folks. 🙌

PR #635, opened 2021-01-15
Add keep_original_cols to some steps

### LINKED ISSUE #282: Suggestion: add a keep_vars arguments (for some steps)

juliasilge:
Closes #282 

This PR adds the long-discussed `keep_original_cols` argument to steps, allowing folks to keep the variables used to compute features in the output of certain steps. Based on the work started by @jonthegeek in #462 at the 2020 tidyverse dev day.

This first set of commits just has `step_pca()`.


R/misc.R
```diff
@@ -637,6 +637,27 @@ check_training_set <- function(x, rec, fresh) {
 }
 
 
+get_keep_original_cols <- function(object) {
```
topepo, line 640: `embed` will need this too. Can you export and use `@keywords internal`?

topepo (changes requested):

R/misc.R
juliasilge: Done ✅

topepo (approved):

Issue #22, opened 2017-01-29
Add steps for basis functions


topepo:
- [x] natural splines
- [x] restricted cubic splines
- [x] orthogonal polynomials
- [x] [periodic splines](https://stat.ethz.ch/R-manual/R-devel/library/splines/html/periodicSpline.html)

alexpghayes:
`mgcv` provides utilities that do a ton of cool basis function calculations that would be cool to incorporate in somewhere.

alexpghayes:
@max Some examples that use periodic splines (although I think the phrase "cyclic splines" is better for Googling). Lots of people use  `s(x, bs = "cc")` in `mgcv`:

- https://peerj.com/preprints/27320/
- https://www.fromthebottomoftheheap.net/2014/05/09/modelling-seasonal-data-with-gam/

tspeidel:
A bit late to the discussion. 
@topepo any plan to add restricted cubic splines? e.g. `tidymodels::step_rcs()` 
They are used a lot and advocated by the likes of Harrell, Steyerberg et al.

PR #640, opened 2021-01-30
Transparently speed up step_bs

### LINKED ISSUE #574: step_bs is extremely slow for large datasets
In practice, especially for modelling tasks, datasets can contain a comparatively small amount of distinct values for predictors - even for continuous.

`splines::bs()` is very inefficient for such predictors, because it computes all basis functions for all rows separately.
This inefficiency is exacerbated by the fact that `recipes:::prep.step_bs` unnecessarily calls `splines::bs()` on the training data, throwing away the most expensive part of the computation altogether.

I have coded a small reprex with a fast preprocessing for `splines::bs()` that yields identical results for all aspects with one small exception: If `knots` is not provided and `df > degree`, the automatic knot selection algorithm for `splines::bs()` will select different knots if called with `unique(x)` instead of `x`.

As such, fixing the performance inefficiency in `recipes` is a little bit (but not much) more work than this proof of concept.

If there is interest in such a performance improvement, I can try coding up a pull request with more efficient versions of `prep.step_bs` and `bake.step_bs`.

### Minimal, runnable code:

``` r
library(recipes)
library(tibble)
set.seed(123)
v_u <- rnorm(100)
x <- tibble(
  v = sample(v_u, 1000000, TRUE)
)

t0 <- Sys.time()
rec <- recipe(x) %>%
  step_bs(v) %>%
  prep()

t1 <- Sys.time()
baked <- rec %>% bake(new_data = x)
t2 <- Sys.time()

bs_fast <- function(x, ...) {
  xu <- unique(x)
  ru <- splines::bs(xu, ...)
  res <- ru[match(x, xu), ]
  copy_attrs <- c("class", "degree", "knots", "Boundary.knots", "intercept")
  attributes(res)[copy_attrs] <- attributes(ru)[copy_attrs]
  res
}

t3 <- Sys.time()
baked_fast <- bs_fast(x$v) %>%
  unclass() %>%
  `attributes<-`(attributes(.)[c("dim", "dimnames")]) %>%
  as_tibble() %>%
  magrittr::set_colnames(sprintf("v_bs_%d", 1L:3L))
t4 <- Sys.time()

identical(baked, baked_fast)
#> [1] TRUE
identical(bs_fast(x$v), splines::bs(x$v))
#> [1] TRUE

cat(glue::glue(
  "prep recipe: {t1 - t0}\n",
  "bake recipe: {t2 - t1}\n",
  "total recipe: {t2 - t0}\n",
  "bake fast: {t4 - t3}"
))
#> prep recipe: 0.714876890182495
#> bake recipe: 0.235795021057129
#> total recipe: 0.950671911239624
#> bake fast: 0.11726713180542
```

<sup>Created on 2020-09-24 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

AshesITR:
fix #574

Testing script for benchmark:

```
devtools::load_all()
library(tibble)
set.seed(123)
v_u <- rnorm(100)
x <- tibble(
  v = sample(v_u, 1000000, TRUE)
)
bench::mark(new_code = recipe(x) %>%
  step_bs(v) %>%
  prep()
)
```

Current master on R 4.0.3, Windows 10 x64, i9-10900K:

```
# A tibble: 1 x 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result   memory                   time         gc              
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>   <list>                   <list>       <list>          
1 old_code      573ms    573ms      1.75     637MB     10.5     1     6      573ms <recipe> <Rprofmem[,3] [693 x 3]> <bch:tm [1]> <tibble [1 x 3]>
```

PR code on same machine:

```
# A tibble: 1 x 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result   memory                     time         gc              
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>   <list>                     <list>       <list>          
1 new_code     71.5ms   71.5ms      14.0     119MB     83.9     1     6     71.5ms <recipe> <Rprofmem[,3] [1,411 x 3]> <bch:tm [5]> <tibble [5 x 3]>
```


AshesITR:
To check that the behaviour didn't change, I checked the objects from both runs using `waldo::compare()`.

```
> waldo::compare(old_result, new_result)
`attr(old$steps[[1]]$terms[[1]], '.Environment')` is <env:00000236F22898F0>
`attr(new$steps[[1]]$terms[[1]], '.Environment')` is <env:00000236FB83C4F8>

`old$steps[[1]]$id`: "bs_yls7H"
`new$steps[[1]]$id`: "bs_Fz6IO"
```

AshesITR:
@juliasilge I can't find the Rout.fail logs to investigate. Can you help me by pointing me where to look?
Thanks!

AshesITR:
cc @juliasilge @topepo 

If you like the implementation, I can include an analogous version for `step_ns` - the knot "prep" code in `splines::ns()` is the same as in `splines::ns()`

AshesITR:
Benchmark results for ns:

```
devtools::load_all()
library(tibble)
set.seed(123)
v_u <- rnorm(100)
x <- tibble(
  v = sample(v_u, 1000000, TRUE)
)
bench::mark(old_code = recipe(x) %>%
  step_ns(v) %>%
  prep()
)
```

```
# A tibble: 1 x 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result   memory                   time         gc              
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>   <list>                   <list>       <list>          
1 new_code      105ms    119ms      6.98     142MB     14.0     4     8      573ms <recipe> <Rprofmem[,3] [876 x 3]> <bch:tm [4]> <tibble [4 x 3]>
```

```
# A tibble: 1 x 13
  expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result   memory                   time         gc              
  <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>   <list>                   <list>       <list>          
1 old_code      657ms    657ms      1.52     977MB     7.61     1     5      657ms <recipe> <Rprofmem[,3] [901 x 3]> <bch:tm [1]> <tibble [1 x 3]>
```

topepo:
Why do we need a function that replicates the `splines` code? Can't we just pre-compute and use the functions as-is? 

For example: 

``` r
library(tidymodels)
#> ── Attaching packages ────────────────────────────────────── tidymodels 0.1.2 ──
#> ✓ broom     0.7.3           ✓ recipes   0.1.15.9000
#> ✓ dials     0.0.9.9000      ✓ rsample   0.0.8.9000 
#> ✓ dplyr     1.0.3           ✓ tibble    3.0.6      
#> ✓ ggplot2   3.3.3           ✓ tidyr     1.1.2      
#> ✓ infer     0.5.3           ✓ tune      0.1.2.9000 
#> ✓ modeldata 0.1.0.9000      ✓ workflows 0.2.1      
#> ✓ parsnip   0.1.4.9000      ✓ yardstick 0.0.7.9000 
#> ✓ purrr     0.3.4
#> ── Conflicts ───────────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard() masks scales::discard()
#> x dplyr::filter()  masks stats::filter()
#> x dplyr::lag()     masks stats::lag()
#> x recipes::step()  masks stats::step()
library(splines)

data(ames)

ames$Sale_Price <- log10(ames$Sale_Price)

set.seed(3)
split <- initial_split(ames)
ames_tr <- training(split)
ames_te <- testing(split)
```

``` r

deg_free <- 10

bs_obj <- bs(ames_tr$Longitude, df = deg_free)
knots <- attributes(bs_obj)$knots
b_knots <- attributes(bs_obj)$Boundary.knots

te_man <- bs(ames_te$Longitude, knots = knots, df = deg_free, Boundary.knots = b_knots)
te_obj <- predict(bs_obj, ames_te$Longitude)
all.equal(te_man, te_obj)
#> [1] TRUE
```

``` r

num_knots <- deg_free - attributes(bs_obj)$degree + 1
knots_manual <- quantile(ames_tr$Longitude, probs = (1:(num_knots-1))/num_knots)
all.equal(knots_manual, knots)
#> [1] TRUE
```

<sup>Created on 2021-02-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>
```

**Comment by AshesITR** [comment]
I'm not sure what you mean.
The point was that directly calling `bs` also evaluates the basis functions, doing a massive useless computation. The PR code omits this part from the `bs` code during prep.

Surely the code in this PR could be refactored a bit to make it more readable, but less similar to the original `bs` code, but directly calling `bs` during prep is the main reason for the poor performance on master.

**Comment by topepo** [comment]
During `prep()`, we can call `quantile()` and `range()` to get the two quantities that we need. That's what the last code block shows (a point I believe that you had previously mentioned). `bs()` and `ns()` would only get called in `bake()`. Do you agree? 

The code in the PR is almost directly copied from the `splines` package and would violate copywrite. We have to make sure that we do not infringe on other people's work or violate the license. I don't see the need for the new function and don't want to make the R Core team a copyright holder for this part of `recipes` if it can be easily avoided.

**Comment by AshesITR** [comment]
Unfortunately, if we want to keep all (including edge-case) behaviour, we'll need to duplicate the logic.

e.g. if you specify `Boundary.knots` such that there are some observations out of range, your reprex starts to fail:

``` r
library(tidymodels)
#> -- Attaching packages -------------------------------------- tidymodels 0.1.2 --
#> v broom     0.7.3      v recipes   0.1.15
#> v dials     0.0.9      v rsample   0.0.8 
#> v dplyr     1.0.3      v tibble    3.0.5 
#> v ggplot2   3.3.3      v tidyr     1.1.2 
#> v infer     0.5.4      v tune      0.1.2 
#> v modeldata 0.1.0      v workflows 0.2.1 
#> v parsnip   0.1.5      v yardstick 0.0.7 
#> v purrr     0.3.4
#> -- Conflicts ----------------------------------------- tidymodels_conflicts() --
#> x purrr::discard() masks scales::discard()
#> x dplyr::filter()  masks stats::filter()
#> x dplyr::lag()     masks stats::lag()
#> x recipes::step()  masks stats::step()
library(splines)

data(ames)

ames$Sale_Price <- log10(ames$Sale_Price)

set.seed(3)
split <- initial_split(ames)
ames_tr <- training(split)
ames_te <- testing(split)

deg_free <- 10

bs_obj <- bs(ames_tr$Longitude, df = deg_free, Boundary.knots = c(-93.6, -93.5))
#> Warning in bs(ames_tr$Longitude, df = deg_free, Boundary.knots = c(-93.6, : some
#> 'x' values beyond boundary knots may cause ill-conditioned bases
knots <- attributes(bs_obj)$knots
b_knots <- attributes(bs_obj)$Boundary.knots

te_man <- bs(ames_te$Longitude, knots = knots, df = deg_free, Boundary.knots = b_knots)
#> Warning in bs(ames_te$Longitude, knots = knots, df = deg_free, Boundary.knots =
#> b_knots): some 'x' values beyond boundary knots may cause ill-conditioned bases
te_obj <- predict(bs_obj, ames_te$Longitude)
#> Warning in bs(x = c(-93.6193873, -93.61732, -93.633826, -93.632852,
#> -93.635951, : some 'x' values beyond boundary knots may cause ill-conditioned
#> bases
all.equal(te_man, te_obj)
#> [1] TRUE

num_knots <- deg_free - attributes(bs_obj)$degree + 1
knots_manual <- quantile(ames_tr$Longitude, probs = (1:(num_knots-1))/num_knots)
all.equal(knots_manual, knots)
#> [1] "Mean relative difference: 0.0005032971"
```

<sup>Created on 2021-02-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


Are we okay with breaking some edge cases?

Regardless, I can definitely rewrite the code to reproduce the same logic assuming a non-error situation if that helps with the copyright problem?

AshesITR:
If we allow for breaking changes and don't want to do the warning / stopping error checks from splines and with the behaviour for ill-conditioned Boundary.knots, we could get away with something like

```
degree <- args$degree %||% switch(type, bs = 3L, ns = 1L)
df <- args$df
intercept <- args$intercept %||% FALSE
boundary <- args$Boundary.knots %||% range(x)

if (!is.null(args$df) && is.null(args$knots)) {
  num_knots <- args$df - degree - intercept
  knots <- quantile(x, (1:(num_knots-1))/num_knots, na.rm = TRUE)
} else {
  knots <- numeric()
}
...
```

would that be okay?

topepo:
> would that be okay?

Yes. 

We already experience the warnings that you show because the test/assessment set data can be out of range (as defined by the training/analysis set). I don't see that as a big issue. 

In your example code, where is the `args` object coming from? I assume that this would be inside of `prep()`.

AshesITR:
Exactly, it would be in `splines_wrapper`, called from both `prep.step_bs` and `prep.step_ns`.

AshesITR:
> We already experience the warnings that you show because the test/assessment set data can be out of range (as defined by the training/analysis set). I don't see that as a big issue.

Except that the new code will, in addition to issuing a warning, pick **different knots** during prep compared to the old code.
The reason is that `splines::bs` computes the quantiles only on the points within `Boundary.knots`, i.e. the old knots were

```
knots <- quantile(x[!is.na(x) & x > boundary[1L] & x < boundary[2L]], probs = ...)
```

where the new ones are

```
knots <- quantile(x, probs = ..., na.rm = TRUE)
```

topepo:
I'm fine with the difference but if you can create code that fulfills the same goal, without using/copying the sources, that would be great. 

I'm not worried about the single row training set edge case so don't go to heroic lengths for that one. You could just add an `rlang::abort()` if that is the case.

topepo:
> Exactly, it would be in `splines_wrapper`, called from both `prep.step_bs` and `prep.step_ns`.

Do we really need that? If/when we extend to other methods, then this needs to be refactored. I'd rather have simpler but separate functions.

AshesITR:
> I'm fine with the difference but if you can create code that fulfills the same goal, without using/copying the sources, that would be great.

Will do.

> Do we really need that? If/when we extend to other methods, then this needs to be refactored. I'd rather have simpler but separate functions.

As you can see in the source, there are only two type-specific branches, one for the default degree (effective default number of knots) and one for the populating the degree attribute or not.

Separating into two functions would create a lot of duplication and the code can hardly be extended to other use-cases, no?
IINM `splines::bs` and `splines::ns` are the only two spline basis steps we have. Have I forgotten one?
I don't feel particularly strongly, though. Do you prefer two near-identical functions?

w.r.t. the bake steps, there might be more use for that outside of this PR, but the implementation is very simple.
I could rename it to `predict_unique`?

juliasilge:
We do have other basis steps we'd like to eventually add like restricted cubic splines and periodic splines, as outlined in #22. Obviously this isn't super high priority as that issue is now 4 years old and we haven't gotten around to it but it would be a reason to consider more separated code, more like how we treat other recipe steps.

AshesITR:
Is this implementation more appealing?
I also added a NEWS bullet as I feel this PR is almost ready for merging. Feel free to change the wording.

AshesITR:
@topepo WDYT about the current state of the PR?
Would you like to see any more changes?

R/ns.R
```diff
@@ -95,21 +95,41 @@ step_ns_new <-
...
+    knots <- numeric()
+  }
+
+  # Only construct the data necessary for splines_predict
+  out <- matrix(NA, ncol = degree + length(knots) + intercept, nrow = 1L)
   class(out) <- c("ns", "basis", "matrix")
-  attr(out, "knots") <- attr(ns_obj, "knots")[]
-  attr(out, "Boundary.knots") <- attr(ns_obj, "Boundary.knots")
-  attr(out, "intercept") <- attr(ns_obj, "intercept")
+  attr(out, "knots") <- knots
```
topepo, line 118: Why are the attributes copied back to the object? These columns are added to the data frame and they are stripped off there.

R/bs.R
```diff
@@ -101,20 +101,41 @@ step_bs_new <-
   }
 
 bs_wrapper <- function(x, args) {
```
topepo, line 103: Please call this `bs_statistics()` since it just does those computations.

topepo (changes requested):
Finally coming around to this repo. Small comments for `bs` and these also apply to `ns`. 

Please add unit tests to show equivalence in the basis function columns produced by these changes with those made by the functions in the `splines` package.

R/ns.R
AshesITR: This can't be removed as they are needed for the predict step.
Commenting out these lines for `bs` errors the tests like this:

```
-- Error (<text>:19:3): correct basis functions --------------------------------
Error: unused arguments (`NA` = NULL, `NA` = NULL, `NA` = NULL, `NA` = NULL)
Backtrace:
 1. recipes::prep(with_bs, training = biomass_tr, verbose = FALSE)
 2. recipes::prep.recipe(with_bs, training = biomass_tr, verbose = FALSE) R/recipe.R:306:8
 4. recipes::bake.step_bs(x$steps[[i]], new_data = training) R/recipe.R:520:8
 5. recipes::bs_predict(...) R/bs.R:174:4
 7. splines:::predict.bs(object, xu)
 8. base::do.call("bs", a)

-- Error (<text>:59:3): printing -----------------------------------------------
Error: unused arguments (`NA` = NULL, `NA` = NULL, `NA` = NULL, `NA` = NULL)
Backtrace:
  1. testthat::expect_output(...)
 11. recipes::prep.recipe(with_bs, training = biomass_tr, verbose = TRUE) R/recipe.R:306:8
 13. recipes::bake.step_bs(x$steps[[i]], new_data = training) R/recipe.R:520:8
 14. recipes::bs_predict(...) R/bs.R:174:4
 16. splines:::predict.bs(object, xu)
 17. base::do.call("bs", a)

Test passed 
```

and `ns` tests fail similarly:

```
-- Error (<text>:19:3): correct basis functions --------------------------------
Error: unused arguments (`NA` = NULL, `NA` = NULL, `NA` = NULL)
Backtrace:
 1. recipes::prep(with_ns, training = biomass_tr, verbose = FALSE)
 2. recipes::prep.recipe(with_ns, training = biomass_tr, verbose = FALSE) R/recipe.R:306:8
 4. recipes::bake.step_ns(x$steps[[i]], new_data = training) R/recipe.R:520:8
 5. recipes::ns_predict(...) R/ns.R:167:4
 7. splines:::predict.ns(object, xu)
 8. base::do.call("ns", a)

-- Error (<text>:59:3): printing -----------------------------------------------
Error: unused arguments (`NA` = NULL, `NA` = NULL, `NA` = NULL)
Backtrace:
  1. testthat::expect_output(...)
 11. recipes::prep.recipe(with_ns, training = biomass_tr, verbose = TRUE) R/recipe.R:306:8
 13. recipes::bake.step_ns(x$steps[[i]], new_data = training) R/recipe.R:520:8
 14. recipes::ns_predict(...) R/ns.R:167:4
 16. splines:::predict.ns(object, xu)
 17. base::do.call("ns", a)

Test passed 
```

R/bs.R
AshesITR: Done.

AshesITR:
The existing tests cover equality for the cases:

https://github.com/tidymodels/recipes/blob/7878d34a0a84212a8b555dd14f12d1b667193bb0/tests/testthat/test_ns.R#L24-L25

https://github.com/tidymodels/recipes/blob/7878d34a0a84212a8b555dd14f12d1b667193bb0/tests/testthat/test_bs.R#L24-L25

Would you like me to add tests for all missing combinations of "missingness" of tuning parameters?

PR #646, opened 2021-02-17
add `allow_rename` arg to `eval_select_recipes()`

hfrick:
This is prep for adding a `step_select()`.

`eval_select_recipes()` still returns a vector of variable names (rather than positions) as positions might differ between training and test sets. If `allow_rename = TRUE`, the returned vector is named with the new names.


R/selections.R
```diff
@@ -94,7 +94,8 @@
 NULL
 
 
-eval_select_recipes <- function(quos, data, info) {
+eval_select_recipes <- function(quos, data, info, allow_rename = FALSE) {
```
DavisVaughan, line 97: Since we plan to export this, and because I would like to match the `tidyselect::eval_select()` interface as much as possible, I think it would be nice to follow tidyselect's lead and add `...` between `info` and `allow_rename = FALSE`.

The dots would have no purpose, except for separating the required arguments from the optional ones. This is nice because it forces you to specify optional arguments by name, which is more robust and provides more information when you are just reading, but not running, the code. We would ensure that the user cannot pass any information in the `...` by calling `ellipsis::check_dots_empty()` as the first line of `eval_select_recipes()`, like tidyselect does. This would require Importing ellipsis, which is fine.

This design guide is full of tips like this
https://design.tidyverse.org/

This particular pattern is data/dots/details in
https://design.tidyverse.org/dots-position.html

See that `eval_select()` uses this pattern (except for the env argument, which is a bit special):

```r
> tidyselect:::eval_select
function (expr, data, env = caller_env(), ..., include = NULL, 
    exclude = NULL, strict = TRUE, name_spec = NULL, allow_rename = TRUE) 
{
    ellipsis::check_dots_empty()
    eval_select_impl(data, names(data), as_quosure(expr, env), 
        include = include, exclude = exclude, strict = strict, 
        name_spec = name_spec, allow_rename = allow_rename)
}
```

R/selections.R
```diff
@@ -106,27 +107,30 @@ eval_select_recipes <- function(quos, data, info) {
...
 
-  # FIXME: Ideally this is `FALSE`, but empty selections incorrectly throw an
+  # FIXME: Ideally this is `FALSE` for strict selection,
+  # but empty selections incorrectly throw an
   # error when this is false due to the following bug:
   # https://github.com/r-lib/tidyselect/issues/221
-  allow_rename <- TRUE
+  # Once it's fixed, remove this and pass allow_rename to
+  # tidyselect::eval_select().
+  allow_rename_ts <- TRUE
```
DavisVaughan, line 116:
```suggestion
  allow_rename_compat <- TRUE
```

Suffixing with `compat`ibility is a nicer indicator to me of what this is for, since now we have a real `allow_rename` argument

R/selections.R
```diff
@@ -106,27 +107,30 @@ eval_select_recipes <- function(quos, data, info) {
...
 
   sel <- tidyselect::eval_select(
     expr = expr,
     data = data,
-    allow_rename = allow_rename
+    allow_rename = allow_rename_ts
```
DavisVaughan, line 121:
```suggestion
    allow_rename = allow_rename_compat
```

R/selections.R
```diff
@@ -106,27 +107,30 @@ eval_select_recipes <- function(quos, data, info) {
...
   )
 
   # Return names not positions, as these names are
   # used for both the training and test set and their positions
-  # may have changed. `sel` won't be named because when `allow_rename = FALSE`,
-  # `eval_select()` returns an unnamed vector.
+  # may have changed. If renaming is allowed, add the new names.
   out <- names(data)[sel]
+  if (allow_rename) names(out) <- names(sel)
```
DavisVaughan, line 128: To be consistent with tidyselect, I think that we should _always_ add `names(sel)` onto `names(out)`. I realize that in the `allow_rename = FALSE` case, this is a bit redundant, but I like that it is all around more consistent.

```r
library(tidyselect)

eval_select(quote(mpg), mtcars)
#> mpg 
#>   1
eval_select(quote(c(new = mpg)), mtcars)
#> new 
#>   1

# for us i think it would look like:

# eval_select_recipes(quote(mpg))
# mpg
# "mpg"

# eval_select_recipes(quote(c(new = mpg)), allow_rename = TRUE)
# new
# "mpg"
```

That will also make it cleaner when the tidyselect bugs finally do get fixed. At that point, the only place that `allow_rename` will be used is where we pass it on to `eval_select()`.

R/selections.R
DavisVaughan, line 97: I debated whether or not `allow_rename` should default to `TRUE` to be more consistent with tidyselect, but I think that it is better to default to `FALSE` as you have it here, since 99% of recipes steps will not allow renaming.

R/selections.R
```diff
@@ -106,27 +107,30 @@ eval_select_recipes <- function(quos, data, info) {
...
 
   # FIXME: Remove this check when the following issue is fixed,
   # i.e. when we can use `allow_rename = FALSE`
```
DavisVaughan, line 131:
```suggestion
  # at that point, just pass `allow_rename` to `eval_select()` directly.
```

tests/testthat/test_rm.R
```diff
@@ -32,13 +34,139 @@ test_that('basics', {
...
+iris_rec <- recipe(~., data = iris)
 
+iris_train <- iris %>%
+  as_tibble() %>%
+  slice(1:75)
+iris_test <- iris %>%
+  as_tibble() %>%
+  # change the position of the variables
+  select(Species, starts_with("Sepal"), starts_with("Petal")) %>%
+  slice(76:150)
```
DavisVaughan, line 46: Consider inlining these global objects into the `test_that()` block that you use them in, or adding a `helper-*.R` file to the `tests/testthat/` directory containing a `generate_iris_test_data()` function that returns a list of these objects.

DavisVaughan (changes requested):

R/selections.R
hfrick: agree - I used _ts for tidyselect but _compat is clearer

tests/testthat/test_rm.R
hfrick: I've put them into the `test_that()` blocks :heavy_check_mark:

R/selections.R
hfrick: done!

R/selections.R
hfrick: Sounds good to me. If we do that, several tests fail though so I'd open a new issue for that and address that in a separate PR.

DavisVaughan: Yea, as discussed offline, we will come back to this elsewhere if we need to!

DavisVaughan (approved):

R/selections.R
hfrick: on a second thought: if consistency with tidyselect is the aim, then maybe always returning a named vector is not what we want after all, given that `eval_select()` returns an unnamed vector when `allow_rename = FALSE`.

``` r
library(tidyselect)

eval_select(quote(mpg), mtcars, allow_rename = FALSE)
#> [1] 1
```

<sup>Created on 2021-03-03 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

DavisVaughan: There is a decent chance that is a bug. I ran across that awhile back https://github.com/r-lib/tidyselect/issues/220

hfrick: Ah, I see :+1:  I've turned your comment into an issue:  #654

PR #655, opened 2021-03-03
be consistent with tidyselect::eval_select()

### LINKED ISSUE #654: Make `eval_select_recipes()` consistent with `tidyselect::eval_select()`
To be consistent with tidyselect, I think that we should _always_ add `names(sel)` onto `names(out)`. I realize that in the `allow_rename = FALSE` case, this is a bit redundant, but I like that it is all around more consistent.

```r
library(tidyselect)

eval_select(quote(mpg), mtcars)
#> mpg 
#>   1
eval_select(quote(c(new = mpg)), mtcars)
#> new 
#>   1

# for us i think it would look like:

# eval_select_recipes(quote(mpg))
# mpg
# "mpg"

# eval_select_recipes(quote(c(new = mpg)), allow_rename = TRUE)
# new
# "mpg"
```

That will also make it cleaner when the tidyselect bugs finally do get fixed. At that point, the only place that `allow_rename` will be used is where we pass it on to `eval_select()`.

_Originally posted by @DavisVaughan in https://github.com/tidymodels/recipes/pull/646#discussion_r582233063_

hfrick:
Closes  #654

`eval_select_recipes()` now returns a vector that is _always_ named, for consistency with `tidyselect::eval_select()`

### DIFF
--- a/R/selections.R
+++ b/R/selections.R
@@ -127,12 +127,12 @@ eval_select_recipes <- function(quos, data, info, ..., allow_rename = FALSE) {
   # used for both the training and test set and their positions
   # may have changed. If renaming is allowed, add the new names.
   out <- names(data)[sel]
-  if (allow_rename) names(out) <- names(sel)
+  names(out) <- names(sel)
 
   # FIXME: Remove this check when the following issue is fixed,
   # at that point, just pass `allow_rename` to `eval_select()` directly.
   # https://github.com/r-lib/tidyselect/issues/221
-  if (!allow_rename & !identical(out, names(sel))) {
+  if (!allow_rename & !identical(unname(out), names(sel))) {
     abort("Can't rename variables in this context.")
   }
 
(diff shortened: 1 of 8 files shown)


R/selections.R
```diff
@@ -127,12 +127,12 @@ eval_select_recipes <- function(quos, data, info, ..., allow_rename = FALSE) {
   # used for both the training and test set and their positions
   # may have changed. If renaming is allowed, add the new names.
   out <- names(data)[sel]
-  if (allow_rename) names(out) <- names(sel)
+  names(out) <- names(sel)
```
DavisVaughan, line 130: It might be cleaner to do something like

```r
names <- names(sel)
```

then in the if statement below:

```r
if (!allow_rename & !identical(out, names)) {
```

then after the if statement

```r
names(out) <- names
```

DavisVaughan (approved):
Yea, even thought this requires a few `unname()`s, I think it is the right thing to do. Thanks for looking at this!

R/selections.R
hfrick: I like that. I had in mind to write it so that we could just delete the block with the if statement when it's time for that - but for the time being this is nicer!

PR #657, opened 2021-03-04
new `step_select()`

### LINKED ISSUE #199: Adding step_select()
It would be nice to have an analogue of `step_rm()` to select variables (much like `dplyr::select()`). If it already exists (I could not find it), it seems meaningful to reference it in the documentation of `step_rm()`. What do you think?

hfrick:
Closes #199 

A new step function to select variables via tidy selectors, role, or type.


R/select.R
```diff
@@ -0,0 +1,155 @@
+#' Select variables using `select`
```
DavisVaughan, line 1: We try to document function names with `()`, so `select()` here
https://style.tidyverse.org/documentation.html#r-code

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+#'
+#' `step_select` creates a *specification* of a recipe step
+#'  that will select variables using [dplyr::select()].
+#'
+#' @inheritParams step_center
+#' @param ... One or more selector functions to choose which
+#'  variables that will be evaluated by the filtering bake. See
```
DavisVaughan, line 8:
```suggestion
#' @param ... One or more selector functions to choose which
#'  variables will be selected when baking. See
```

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+#' rec_train <- bake(prepped, new_data = NULL)
+#' all.equal(dplyr_train, rec_train)
+#'
+#' dplyr_test <-
+#'   iris %>%
+#'   as_tibble() %>%
+#'   slice(76:150) %>%
+#'   select(Species, starts_with("Sepal"))
+#' rec_test <- bake(prepped, iris %>% slice(76:150))
+#' all.equal(dplyr_test, rec_test)
```
DavisVaughan, line 52: I think this can be a little more compactly written as:

```r
library(dplyr)

iris_tbl <- as_tibble(iris)
iris_train <- slice(iris_tbl, 1:75)
iris_test <- slice(iris_tbl, 76:150)

dplyr_train <- select(iris_train, Species, starts_with("Sepal"))
dplyr_test <- select(iris_test, Species, starts_with("Sepal"))

rec <- recipe(~., data = iris_train) %>%
  step_select(Species, starts_with("Sepal")) %>%
  prep(training = iris_train)

rec_train <- bake(rec, new_data = NULL)
all.equal(dplyr_train, rec_train)

rec_test <- bake(rec, iris_test)
all.equal(dplyr_test, rec_test)
```

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+#' bake(qq_rec, new_data = NULL) %>% slice(1:4)
+#'
+#' # The difference:
+#' tidy(qq_rec, number = 1)
+#' tidy(qq_rec, number = 2)
+step_select <- function(recipe,
+                        ...,
+                        role = NA,
+                        trained = FALSE,
+                        inputs = NULL,
```
DavisVaughan, line 74: I don't _think_ we need `inputs` as an argument here. I imagine that internally we can just reuse the `$terms` slot to hold the selected character vector of terms after calling `eval_select_recipes()`. The user would never use `inputs` here, and it would be overwritten even if they did

See below for some edits about how I'd do this

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+      trained = trained,
+      inputs = inputs,
+      skip = skip,
+      id = id
+    )
+  }
+
+#' @export
+prep.step_select <- function(x, training, info = NULL, ...) {
+  sel <- eval_select_recipes(x$terms, training, info, allow_rename = TRUE)
```
DavisVaughan, line 103:
```suggestion
  terms <- eval_select_recipes(x$terms, training, info, allow_rename = TRUE)
```

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+
+  step_select_new(
+    terms = x$terms,
+    trained = TRUE,
+    role = x$role,
+    inputs = sel,
```
DavisVaughan, line 109: Then you could remove this line

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+    skip = x$skip,
+    id = x$id
+  )
+}
+
+#' @export
+bake.step_select <- function(object, new_data, ...) {
+  dplyr::select(new_data, object$inputs)
```
DavisVaughan, line 117:
```suggestion
  dplyr::select(new_data, dplyr::all_of(object$terms))
```

Two changes:

- Use `$terms` now rather than `$inputs`

- I wasn't aware that `select()` allows a named character vector as input. That doesn't look to be documented behavior, and the blessed way to do this is to use `all_of()` (as mentioned above in the examples section) so you might consider this way instead

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+    }
+    invisible(x)
+  }
+
+#' @rdname step_select
+#' @param x A `step_select` object
+#' @export
+tidy.step_select <- function(x, ...) {
+  if (is_trained(x)) {
+    res <- tibble(terms = x$inputs)
```
DavisVaughan, line 147: This would change to `$terms`, but you might also consider unnaming it? Or using the `names()` here instead, if you want to show the renamed results instead of what was selected - up to you

tests/testthat/test-select.R
```diff
@@ -0,0 +1,205 @@
...
+
+  if (r_version_pre_36) {
+    suppressWarnings(RNGversion("3.5.0"))
+  }
+
+  set.seed(403)
+  petal <- c("Petal.Width", "Petal.Length")
+  rec <- recipe(~., data = iris) %>%
+    step_select(species = Species, starts_with("Sepal"), petal) %>%
+    step_select(!!petal)
```
DavisVaughan, line 189: I think it would be less mental work to just fix the `id` here in these two steps. So `step_select(id = "select_no_qq")` or something for the first one. Then you don't need to worry about the rngkind. Sorry if we confused you about this in the group meeting

DavisVaughan (changes requested):

R/select.R
hfrick: Nice, thanks!

R/select.R
```diff
@@ -0,0 +1,155 @@
...
+#'   step_select(Species, sepal_vars) %>%
+#'   # best approach
+#'   step_select(Species, !!sepal_vars) %>%
+#'   prep(training = iris)
+#'
+#' bake(qq_rec, new_data = NULL) %>% slice(1:4)
+#'
+#' # The difference:
+#' tidy(qq_rec, number = 1)
+#' tidy(qq_rec, number = 2)
```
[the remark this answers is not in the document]
hfrick: oh good catch, thanks! yes, I was trying to show how to best use quasi-quotation (similar to what I saw in the examples of `step_mutate()`)

R/select.R
hfrick: :+1: I've changed it to use only `$terms` and removed `$inputs`

R/select.R
hfrick: I went with the (new) names :+1:

R/select.R
hfrick: went with `all_of()`!

tests/testthat/test-select.R
```diff
@@ -0,0 +1,205 @@
...
+  rec_test <- bake(prepped, iris_test)
+  expect_equal(dplyr_test, rec_test)
+})
+
+test_that("quasiquotation", {
+  sepal_vars <- c("Sepal.Width", "Sepal.Length")
+
+  rec_1 <-
+    recipe(~., data = iris) %>%
+    step_select(Species, sepal_vars)
```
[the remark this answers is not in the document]
hfrick: `all_of()` it is!

tests/testthat/test-select.R
hfrick: yeah, that makes it cleaner! :raised_hands:

R/select.R
hfrick: I spotted a few things which weren't consistent across dyplr-based steps --> #660

R/select.R
```diff
@@ -0,0 +1,155 @@
+#' Select variables using `select`
+#'
+#' `step_select` creates a *specification* of a recipe step
```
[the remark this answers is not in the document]
hfrick: Most steps are not yet documented like this, I've opened another issue for that: #661

R/select.R
hfrick: I've used this pattern also in the tests now

Issue #652, opened 2021-02-23
The rentain flag in prep has weird behaviour


lukasal:
## The problem

The retain flag in prep has weird behavior! The $template variable in the prepped recipe is never empty!
Either the retain flag is FALSE and the data from the recipe definition is passed, or the flag is TRUE and the data from the prep function is passed to $template. This is very unexpected as I would like to have no training data in case of retain = FALSE!

In the first possibility (retain = FALSE), 
https://github.com/tidymodels/recipes/blob/d8858f8aedd10cfccb620c586d26c61632d0ea1d/R/recipe.R#L209-L216
the $template variable is filled during the recipe definition and not altered during the prep routine.
In case of retain = TRUE it is overriden when training data is specified.

https://github.com/tidymodels/recipes/blob/d8858f8aedd10cfccb620c586d26c61632d0ea1d/R/recipe.R#L479-L486

My expectation would be to change that code to 

``` r
if (retain) {
      if (verbose)
        cat("The retained training set is ~",
            format(object.size(training), units = "Mb", digits = 2),
            " in memory.\n\n")

      x$template <- training
    } else {
      x$template <- NULL
    }
``` 
For me this is important as I have a dataset of multiple GB which I don't want to keep for later, when I use the model! I'm happy to fix that and create a pull request if there is agreement on the fix.

## Reproducible example

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
    library(recipes)
#> 
#> Attaching package: 'recipes'
#> The following object is masked from 'package:stats':
#> 
#>     step
    
    train_mtcars <- mtcars %>% {do.call("rbind", replicate(10000, ., simplify = FALSE))}
    # Error if I dont specify a dataset here:
    mtcars_recipe <- recipe(mpg ~ hp + wt)
#> Error in form2args(formula, data, ...): argument "data" is missing, with no default
    
    mtcars_recipe <- recipe(mpg ~ hp + wt, data = train_mtcars)
    
    v1 <- prep(mtcars_recipe, retain = TRUE, verbose = TRUE)
#> The retained training set is ~ 7.33 Mb  in memory.
    v2 <- prep(mtcars_recipe, retain = FALSE, verbose = TRUE)
    
    print(object.size(v1))
#> 7692800 bytes
    print(object.size(v2))
[...]

PR #658, opened 2021-03-08
retain flag behaviour fixed Issue #652

lukasal:
Attempt to fix the bug (issue https://github.com/tidymodels/recipes/issues/652) for the retain flag, including test to ensure correct functionality of the retain flag.


topepo:
It looks great! Can you add an item to the NEWS file?

PR #680, opened 2021-04-05
Fix variable name in example

mikemc:


juliasilge:
Thank you so much @mikemc! 🙌

PR #685, opened 2021-04-14
Turn off parallel processing for NMF, do not load pkgs

### LINKED ISSUE #681: Loading the required packages for `step_nnmf()` makes `fit()` misbehave
I was working on tidymodels/extratests#23 to move the tests for `step_nnmf()` out of recipes here (they are super long) when I noticed [some pretty strange errrors](https://github.com/tidymodels/extratests/runs/2272764312).

``` r
library(tidymodels)
#>   method                   from   
#>   required_pkgs.model_spec parsnip
data(biomass)

rec <- recipe(HHV ~ ., data = biomass) %>%
  update_role(sample, new_role = "id var") %>%
  update_role(dataset, new_role = "split variable") %>%
  step_nnmf(all_numeric_predictors(), num_comp = 2, seed = 473, num_run = 2) %>%
  prep(training = biomass)

linear_reg() %>% fit(mpg ~ ., data = mtcars)
#> Error in (function (classes, fdef, mtable) : unable to find an inherited method for function 'fit' for signature '"linear_reg"'
```

<sup>Created on 2021-04-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

This error is happening because for some reasons `NMF::fit` is being called instead of `parsnip::fit`.

The required packages for this recipes step are loaded via `require()` within `prep()`, and once NMF is loaded, it seems to be causing some havoc:

https://github.com/tidymodels/recipes/blob/8152d8ec2419523f92b8217f09b99c70ec0744fe/R/nnmf.R#L150-L154

I am guessing from this that nobody has ever tried using this recipe step in a modeling context?

juliasilge:
Closes #681 

This PR turns off the parallelization that NMF has on by default, which allows us to use NMF and dimRed without fully loading them as before. This solves the problem with `fit()`:

``` r
library(tidymodels)
#>   method                   from   
#>   required_pkgs.model_spec parsnip
data(biomass)

rec <- recipe(HHV ~ ., data = biomass) %>%
  update_role(sample, new_role = "id var") %>%
  update_role(dataset, new_role = "split variable") %>%
  step_nnmf(all_numeric_predictors(), num_comp = 2, seed = 473, num_run = 2) %>%
  prep(training = biomass)

linear_reg() %>% fit(mpg ~ ., data = mtcars)
#> Warning: Engine set to `lm`.
#> parsnip model object
#> 
#> Fit time:  3ms 
#> 
#> Call:
#> stats::lm(formula = mpg ~ ., data = data)
#> 
#> Coefficients:
#> (Intercept)          cyl         disp           hp         drat           wt  
#>    12.30337     -0.11144      0.01334     -0.02148      0.78711     -3.71530  
#>        qsec           vs           am         gear         carb  
#>     0.82104      0.31776      2.52023      0.65541     -0.19942
```

<sup>Created on 2021-04-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

In the future, we could consider exposing some kind of parallel option for `step_nnmf()` for folks who use it outside of our tuning infrastructure and want to run it in parallel.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,7 +17,7 @@
 
 * Fixed behavior of the retain flag in `prep()` (#652).
 
-* The `tidy()` methods for `step_nnmf()` was rewritten since it was not great. (#665)
+* The `tidy()` methods for `step_nnmf()` was rewritten since it was not great (#665), and `step_nnmf()` now no longer fully loads underlying packages (#685). 
 
 ## Improvements and Other Changes
 
--- a/R/nnmf.R
+++ b/R/nnmf.R
@@ -22,7 +22,8 @@
 #'  to obtain a consensus projection.
 #' @param options A list of options to `nmf()` in the NMF package by way of the
 #'  `NNMF()` function in the `dimRed` package. **Note** that the arguments
-#'  `data` and `ndim` should not be passed here.
+#'  `data` and `ndim` should not be passed here, and that NMF's parallel
+#'  processing is turned off in favor of resample-level parallelization.
 #' @param res The `NNMF()` object is stored
 #'  here once this preprocessing step has been trained by
 #'  [prep.recipe()].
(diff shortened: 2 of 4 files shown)


DavisVaughan (approved):
Yay!

PR #686, opened 2021-04-15
Typo: their - > there

schoonees:
Corrected a typo in vignette: their -> there


juliasilge (approved):
Thank you so much @schoonees! 🙌

PR #701, opened 2021-04-27
Move all steps' `tidy` docs

### LINKED ISSUE #695: Removing `tidy.*()` from Github Pages
Similar to https://github.com/tidymodels/parsnip/issues/423

The presence of `tidy.*()` with every step adds a lot of noise

# Old reference

Screen Shot 2021-04-23 at 11 53 33 AM

# Proposed reference

Screen Shot 2021-04-23 at 11 56 02 AM

https://recipes.tidymodels.org/reference/index.html

juliasilge:
Closes #695

This PR moves the documentation for `tidy` methods from the individual recipe step/check pages to the already-existing centralized `tidy` page.

Is there a way to change the order of these so that this isn't located at "step_BoxCox"?


DavisVaughan:
> Is there a way to change the order of these so that this isn't located at "step_BoxCox"?

I don't think so. I think this is some bad interaction between the `@name` tag, which isn't getting respected, and the fact that we are forcibly showing an S3 method `tidy.step_BoxCox` on the man page. For some reason that is overriding the `@name` tag, and since that one is the first alphabetically, it shows up in the man page as `\name{tidy.step_BoxCox}`.

Note that we do still get an alias for `tidy.recipe`, i.e. `\alias{tidy.recipe}` in the man file, but the actual name of the file in the top left corner of the man page still shows `tidy.step_BoxCox`, which feels like a roxygen bug.

I have the same issue in clock. i.e. `` ?clock::`clock-arith` `` points to the right page, but the name in the top left corner is `vec_arith.clock_year_day`, which is the first method alphabetically.

DavisVaughan (approved):
Whew!

Issue #689, opened 2021-04-16
`step_modeimpute` returns all NA when column is completely null, instead of mode from training data


jake-mason:
I'm not sure if this is intended behavior, or if I'm missing some other step that could help me to avoid this issue, but when using a recipe with `step_modeimpute` (this may also apply to other steps – didn't verify) on a factor column, when the column in `new_data` is completely NA, NAs are returned, as opposed to the mode from the training dataset.

```R
df <- data.frame(x = as.factor(c(0, 1, 0)))
binary_features <- 'x'

base_recipe <- recipes::recipe(df) %>%
	update_role(all_of(binary_features), new_role = 'predictor') %>%
	# Impute mode for binary features
	step_modeimpute(one_of(binary_features))

prepped_recipe <- prep(
	base_recipe,
	training = df,
	strings_as_factors = FALSE,
	retain = FALSE,
	fresh = TRUE
)
```

Let's try it out:

```
# Our original data.frame – looks good!
> bake(prepped_recipe, new_data = df)
# A tibble: 3 x 1
  x    
  <fct>
1 0    
2 1    
3 0    
# A new data.frame, with NA, 1, and 0 present. Still looks good – the mode is imputed.
> bake(prepped_recipe, new_data = data.frame(x = as.factor(c(NA, NA, 1, 0))))
# A tibble: 4 x 1
  x    
  <fct>
1 0    
2 0    
3 1    
4 0    
# Another data.frame, this time completely NA. No mode imputed.
> bake(prepped_recipe, new_data = data.frame(x = as.factor(c(NA, NA))))
# A tibble: 2 x 1
  x    
  <fct>
1 NA   
2 NA   
```

This behavior seems strange to me, but then again maybe I'm missing something and there's good reason for this.

juliasilge:
This is happening because of how `recipes:::cast()` works:

``` r
recipes:::cast("0", as.factor(c(NA, NA, 1, 0)))
#> [1] 0
#> Levels: 0 1
recipes:::cast("0", as.factor(c(NA, NA)))
#> [1] <NA>
#> Levels:
```

<sup>Created on 2021-04-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

DavisVaughan:
So I took a closer look at this. I think the problem is not necessarily with `cast()`, but with the data that we are passing to `cast()`. This is a case where:

- At training time we have a factor with a set of levels
- At test time we have a factor with a _different_ set of levels

[...]

PR #704, opened 2021-05-06
Added test for mode imputation

jake-mason:
@juliasilge here's a test for the mode imputation steps that aligns with the examples we went through in the issue here: https://github.com/tidymodels/recipes/issues/689. Lemme know how else I can help.


juliasilge:
Thank you so much for your contributions on this issue @jake-mason! Closed in #719

Issue #287, opened 2019-02-10
id fields converted to NA


topepo:
From [this post](https://community.rstudio.com/t/how-to-specify-a-column-to-be-unaffected-in-recipes/23056/6). 

``` r
library(dplyr, warn.conflicts = FALSE)
library(recipes, warn.conflicts = FALSE)

data <- tibble(
  id = letters[1:12],
  output = rnorm(12, mean = 0),
  pred1 = rnorm(12, mean = 10),
  pred2 = rnorm(12, mean = 20),
  pred3 = factor(rep(c('f1', 'f2', 'f3'), 4))
)

data$pred1[c(1,6)] <- NA
data$pred2[c(2,7)] <- NA
df_train <- data[1:5,]
df_test <- data[6:10,]

rec_obj <- recipe(x = df_train) %>%
  update_role(output, new_role = 'outcome') %>%
  update_role(id, new_role = "id variable") %>%
  update_role(-output, -id, new_role = 'predictor') %>%
  step_dummy(pred3) %>%
  step_center(pred1, pred2) %>%
  step_scale(pred1, pred2) %>%
  step_medianimpute(all_predictors())

rec_trained <- prep(rec_obj, training = df_train)
train_data    <- bake(rec_trained, new_data = df_train)
test_data     <- bake(rec_trained, new_data = df_test)
test_data
#> # A tibble: 5 x 6
#>   id     output  pred1  pred2 pred3_f2 pred3_f3
#>   <fct>   <dbl>  <dbl>  <dbl>    <dbl>    <dbl>
#> 1 <NA>  -1.66    0.155 -0.484        0        1
#> 2 <NA>  -0.0917 -1.63  -0.375        0        0
#> 3 <NA>   0.0398  0.143 -0.975        1        0
#> 4 <NA>   0.0514  0.612  0.110        0        1
#> 5 <NA>  -0.746  -2.95  -2.93         0        0
```

Created on 2019-02-10 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)

DavisVaughan:
I'm not sure this is a bug. I think that user just wanted to use `strings_as_factors = FALSE`. 

It looks like when `strings_as_factors` is `TRUE`, then in `prep()` all character columns are coerced to factor and the levels are stored (so this happened to `id` and the levels `a` through `e` were stored). 

[...]

PR #706, opened 2021-05-12
strings_as_factors in recipe

### LINKED ISSUE #331: strings_as_factors
This is an option to `prep()` but it is really intrinsic to the recipe so it should probably an argument there. We can soft-deprecate the `prep()` argument.

smingerson:
Closes #331. 

As mentioned in #287, this is important to be able to control how a recipe behaves when used within a workflow -- as far as I can tell, it is impossible to change it from `strings_as_factors=TRUE`, even using sketchy methods like [namespace modification](https://coolbutuseless.github.io/2018/04/11/changing-the-default-arguments-to-a-function/).

Let me know what changes are needed.


juliasilge:
Thank you so much for this contribution @smingerson! 🙌 As you know from #331 this is something we definitely want implemented. There are likely going to be some changes that either we add here to your PR or that we talk about with you; we have various folks going on vacation so I wanted to make sure we touched base before your PR got too old.

EmilHvitfeldt:
Outline of what Happens in this PR:

- Added an `strings_as_factors` argument to `recipe.data.frame()` with the default `NULL`.
- `recipe.matrix()` and `recipe.formula()` gets access to `strings_as_factors` as the dots are passed from those functions to `recipe.data.frame()`.
- The `strings_as_factors` argument stays in `prep.recipe()` with the default `TRUE` and the added `strings_as_factors` has `NULL` as its default.
    - The `NULL` allows us to easily check whether `recipe()` or `prep()` set the argument.
    - If `strings_as_factors` from `recipe()` is not `NULL` it will win no matter what.

We have a lifecycle warning if someone sets `strings_as_factors` in `prep()`. This will likely cause a little friction, but should be easy to come around, especially since i don't expect many people to save a un-prepped recipe for prepping, when you could just as easily save the code to create it.

EmilHvitfeldt:
# Revdepcheck

Ran clean 🎉 

# Related issues

These issues have been  tested against this PR and passed

- https://github.com/tidymodels/recipes/issues/287

These issues have been  tested against this PR and passed

- None

tests/testthat/test-basics.R
```diff
@@ -532,3 +533,91 @@ test_that("recipe() error for table input (#1416)", {
     recipe(Titanic, Survived ~ .)
   )
 })
+
+
```
EmilHvitfeldt, line 537: This is where we are actually testing this behavior. All the other changes to the test files are moving the arg from `prep()` to `recipe()`

topepo:
@smingerson thanks for doing this and apologies for letting it linger for so long

PR #723, opened 2021-06-14
Geo latlon

jkennel:
- `step_depth`  typo.

- `step_geodist` used Pythagorean theorem which is misleading for latitude and longitude.  The method currently allows for one ref_lon and ref_lat so the use of `apply` was computationally disadvantageous in the previous code.  If the goal is to allow for multiple ref_lon and ref_lat, distance matrices, or bearing I can create another pull request.
- `step_geodist` Added the Haversine function for great circle distances.  You may want to switch to a geospatial package if you want other methods, but this should cover many common cases.
- `step_geodist` The Haversine or Pythagorean is chosen through a new argument is_lat_lon.

### DIFF
--- a/R/geodist.R
+++ b/R/geodist.R
@@ -12,6 +12,9 @@
 #' @param role or model term created by this step, what analysis
 #'  role should be assigned?. By default, the function assumes
 #'  that resulting distance will be used as a predictor in a model.
+#' @param is_lat_lon A logical: Are coordinates in latitude and longitude? If
+#'  TRUE the Haversine formula is used and the returned result is meters. If
+#'  FALSE the Pythagorean formula is used. Default is TRUE.
 #' @param log A logical: should the distance be transformed by
 #'  the natural log function?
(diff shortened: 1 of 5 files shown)


R/geodist.R
```diff
@@ -60,6 +66,8 @@ step_geodist <- function(recipe,
     rlang::abort("`ref_lon` should be a single numeric value.")
```
topepo, line 66: Can you add some logic to determine if existing objects have `is_lat_lon` and, if not, set to FALSE? We don't want to break existing recipes.

topepo (commented):
Looks great. Thanks.

Can you add an entry to the NEWS file too?

topepo:
Any reason to delete this and close the issue?

jkennel:
I deleted it because I found the posting guidelines.  It said not to submit a pull request unless you first discussed it as an issue to see if it was a desired change. I can make the updates and submit today.

PR #727, opened 2021-06-15
pr 726 (memory and speed improvement)

jkennel:
When the terms attributes are desired from `model.frame` a subset of the data can be used to improve speed and memory use particularly for large datasets.


topepo:
Can you add a note in the NEWS file about this?

topepo:
Thanks again

PR #728, opened 2021-06-15
pr 725 step_geodist calculates great-circle distance for lat-lon pairs

jkennel:
When coordinates are in lat-lon distances are calculated using the Haversine formula.  Otherwise they use the Pythagorean formula.


topepo:
Thanks

PR #734, opened 2021-06-24
Specify which step produces error with withCallingHandlers()

### LINKED ISSUE #420: Specify which step produces error
Not all steps report which step they are produced inside. It can lead to slowness when debugging. This is a more general issue that contains #394. 
``` r
library(recipes)

recipe(~ ., data = iris) %>%
  step_dummy(Sepal.Length) %>%
  step_scale(all_predictors()) %>%
  prep()
#> Warning: The following variables are not factor vectors and will be ignored:
#> `Sepal.Length`
#> Error: The `terms` argument in `step_dummy` did not select any factor columns.

recipe(~ ., data = iris) %>%
  step_discretize(Species) %>%
  step_scale(all_predictors()) %>%
  prep()
#> Error: All columns selected for the step should be numeric
```

<sup>Created on 2019-12-01 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

EmilHvitfeldt:
This PR tries to close #420 by using `withCallingHandlers()` to modify the error message to include the name of the step which produces the error. Specifically when the error happens when the step is `prep()`ed in `prep.recipe()`.

I would like some feedback on the error message modification as well as I'm not 100% happy with it.

Pro: It appears to work where I have tested it.

Con: The traceback gets slightly more complicated. See `11:` and `4:` below.

``` r
library(recipes)

recipe(~ ., data = iris) %>%
  step_scale(all_predictors()) %>%
  prep()
#> Error: [in step_scale()] All columns selected for the step should be numeric
```

Traceback:

``` r
12: stop(c) at recipe.R#446
11: (function (c) 
    {
        c$message <- paste0("[in ", attr(x$steps[[i]], "class")[1], 
            "()] ", c$message)
        stop(c)
    })(structure(list(message = "All columns selected for the step should be numeric", 
        trace = structure(list(calls = list(recipe(~., data = iris) %>% 
            step_scale(all_predictors()) %>% prep(), recipes::prep(.), 
            recipes::prep.recipe(.), base::withCallingHandlers(expr = prep(x$steps[[i]], 
                training = training, info = x$term_info), error = function(c) {
                c$message <- paste0("[in ", attr(x$steps[[i]], "class")[1], 
                    "()] ", c$message)
                stop(c)
            }), recipes::prep(x$steps[[i]], training = training, 
                info = x$term_info), recipes::prep.step_scale(x$steps[[i]], 
                training = training, info = x$term_info), recipes::check_type(training[, 
                col_names])), parents = c(0L, 0L, 0L, 3L, 3L, 3L, 
        6L), indices = 1:7), class = "rlang_trace", version = 1L), 
        parent = NULL), class = c("rlang_error", "error", "condition"
    )))
10: signalCondition(cnd)
9: signal_abort(cnd)
8: rlang::abort(paste0("All columns selected for the step", " should be ", 
       label)) at misc.R#427
7: check_type(training[, col_names]) at scale.R#107
6: prep.step_scale(x$steps[[i]], training = training, info = x$term_info) at recipe.R#306
5: prep(x$steps[[i]], training = training, info = x$term_info) at recipe.R#439
4: withCallingHandlers(expr = prep(x$steps[[i]], training = training, 
       info = x$term_info), error = function(c) {
       c$message <- paste0("[in ", attr(x$steps[[i]], "class")[1], 
           "()] ", c$message)
       stop(c)
   }) at recipe.R#439
3: prep.recipe(.) at recipe.R#306
2: prep(.)
1: recipe(~., data = iris) %>% step_scale(all_predictors()) %>% 
       prep()
```

This is the traceback for the same error without the fix

``` r
9: stop(fallback)
8: signal_abort(cnd)
7: rlang::abort(paste0("All columns selected for the step", " should be ", 
       label)) at misc.R#427
6: check_type(training[, col_names]) at scale.R#107
5: prep.step_scale(x$steps[[i]], training = training, info = x$term_info) at recipe.R#306
4: prep(x$steps[[i]], training = training, info = x$term_info) at recipe.R#439
3: prep.recipe(.) at recipe.R#306
2: prep(.)
1: recipe(~., data = iris) %>% step_scale(all_predictors()) %>% 
       prep()
```

`check_type()`s error message should be modified if this PR goes through as it is half-baked right now

https://github.com/tidymodels/recipes/blob/a2155c52193a8238a5b06cbd2ab34bade39684ad/R/misc.R#L428-L433

### DIFF
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -440,7 +440,7 @@ prep.recipe <-
           expr = prep(x$steps[[i]], training = training, info = x$term_info),
           error = function(c) {
             c$call <- call(attr(x$steps[[i]], "class")[1])
-            stop(c)
+            rlang::cnd_signal(c)
           }
         )
 
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -439,9 +439,7 @@ prep.recipe <-
         x$steps[[i]] <- withCallingHandlers(
           expr = prep(x$steps[[i]], training = training, info = x$term_info),
           error = function(c) {
-            c$message <- paste0(
-              "[in ", attr(x$steps[[i]], "class")[1], "()] ",c$message
-            )
+            c$call <- call(attr(x$steps[[i]], "class")[1])
             stop(c)
           }
         )
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -436,10 +436,16 @@ prep.recipe <-
 
         # Compute anything needed for the preprocessing steps
         # then apply it to the current training set
-        x$steps[[i]] <-
-          prep(x$steps[[i]],
-               training = training,
-               info = x$term_info)
+        x$steps[[i]] <- withCallingHandlers(
+          expr = prep(x$steps[[i]], training = training, info = x$term_info),
+          error = function(c) {
+            c$message <- paste0(
+              "[in ", attr(x$steps[[i]], "class")[1], "()] ",c$message
+            )
+            stop(c)
+          }
+        )
+
         training <- bake(x$steps[[i]], new_data = training)
         x$term_info <-
           merge_term_info(get_types(training), x$term_info)


EmilHvitfeldt:
This approach could be applied to warnings and messages as well if I recall correctly.

DavisVaughan:
I think I would much rather add our own `abort_step(step_name, message)` helper that we would call as `abort_step("step_dummy", "All columns should be numeric")` and then use it systematically in each step, passing through the `step_name ` string argument down to any helpers like `check_type()` as required.

In vctrs we use the convention of setting the default for things like this to the empty string, i.e. `step_name = ""`. If we see the empty string, then we'd just report the error message without the step name information. This would allow us to add this argument to exported helpers like `check_type()` in a backwards compatible way, since it would make the extra step name info optional.

I mainly worry that catching and rethrowing errors could have unintended side effects (more complicated tracebacks being one of them). If it is possible to avoid doing it this way, then my personal preference is to do so

EmilHvitfeldt:
That is good feedback!

Correct me if I'm wrong. But does that approach require that input that would cause errors being caught and aborted before executing? More specifically, would your approach be able to handle the following error?

``` r
library(recipes)
set.seed(1234)

Sample <- matrix(rnorm(30 * 5000), 30)
Sample[, c(128, 256, 512)] <- 1

recipe(~., data = as.data.frame(Sample)) %>%
  step_normalize(all_predictors()) %>%
  step_pca(all_predictors()) %>%
  prep()
#> Error in svd(x, nu = 0, nv = k): infinite or missing values in 'x'
```

DavisVaughan:
That's a good point. It won't catch that. We may have to use your approach. I'll ask Lionel if there is anything we need to be careful about with this and will get back to you on it tomorrow

lionel-:
You could use the `call` field to indicate the call. Unlike `stop()`'s default, this shouldn't include any arguments, just a bare call like `quote(step_dummy()`. This way R will mention the call itself and you don't need to modify the message.

@hadley is now thinking about including the call (stripped from argument) by default in `abort()`. Maybe we could also consult a global option like `rlang_abort_call` that could be set dynamically (withr-like). We could then set this option in the main dplyr, tidyr, ggplot2, and tidymodels verbs. Only if unset would we look at the immediate frame call. With this approach the main purpose of displaying calls would be to inform users of pipelines which step did the error occur in.

DavisVaughan:
Is there any way to do this while preserving the niceties of rlang errors? Like the footer encouraging you to look at `rlang::last_error()`?

(Side note that `reprex::reprex()` didn't work nicely with this, and I've had to construct the below reprex manually by running it interactively and copying the output over. I couldn't figure out the right combination of global options to set to get it to show me the same output in a reprex)

```r
enhancer <- function(expr, step_name) {
  withCallingHandlers(
    expr = expr, 
    error = function(cnd) {
      cnd$call <- call(step_name)
      stop(cnd)
    })
}

aborter <- function() {
  rlang::abort("oh no")
}

# Not great
enhancer(aborter(), "step_dummy")
#> Error in step_dummy() : oh no

rlang::last_error()
#> Error: Can't show last error because no error was recorded yet

# This is the nice stuff we lost
aborter()
#> Error: oh no
#> Run `rlang::last_error()` to see where the error occurred.

rlang::last_error()
#> <error/rlang_error>
#> oh no
#> Backtrace:
#>  1. global::aborter()
#> Run `rlang::last_trace()` to see the full context.
```

lionel-:
Yes we'd fix that.

EmilHvitfeldt:
Alright, I switched to `c$call` and it appears to work

``` r
library(recipes)

recipe(~ ., data = iris) %>%
  step_scale(all_predictors()) %>%
  prep()
#> Error in step_scale(): All columns selected for the step should be numeric

set.seed(1234)

Sample <- matrix(rnorm(30 * 5000), 30)
Sample[, c(128, 256, 512)] <- 1

recipe(~., data = as.data.frame(Sample)) %>%
  step_normalize(all_predictors()) %>%
  step_pca(all_predictors()) %>%
  prep()
#> Error in step_pca(): infinite or missing values in 'x'
```

<sup>Created on 2021-07-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

R/recipe.R
```diff
@@ -436,10 +436,14 @@ prep.recipe <-
...
         # then apply it to the current training set
-        x$steps[[i]] <-
-          prep(x$steps[[i]],
-               training = training,
-               info = x$term_info)
+        x$steps[[i]] <- withCallingHandlers(
+          expr = prep(x$steps[[i]], training = training, info = x$term_info),
+          error = function(c) {
+            c$call <- call(attr(x$steps[[i]], "class")[1])
+            stop(c)
```
lionel-, line 443: This should use `rlang::cnd_signal()` instead of `stop()` to preserve the reminder to call `last_error()`. With the released version of rlang this will not actually display the `call` field. However in the next version it will, see https://github.com/r-lib/rlang/pull/1235.

EmilHvitfeldt: Thank you!

DavisVaughan:
FYI we might choose a different way to handle this - rather than modifying the original error condition, we might create a completely new error condition specific to recipes that has an error message that would mention the step name, and would set the original error condition as the _parent_ of the new recipes specific error condition. Something like:

```r
abort(wrapper_msg, call = call("step_foo"), parent = original_cnd)
```

DavisVaughan:
@EmilHvitfeldt I appreciate you working on this PR, I think we will wait for the next version of rlang to come out, and then I'll tackle this one myself using some of the new changes!

PR #735, opened 2021-06-24
Issue/702 add step_harmonic

jkennel:
This pull request adds `step_harmonic`. `step_harmonic` generates sin and cos terms that can be used to model cyclical components. I was somewhat unclear on the optimal `tidy` (what to include) and `tunable` (tuning on period/frequency) methods so they could use a review.

In addition, there are two minor fixes for other steps:

- `step_date` error message was not outputting possible feature values. 
- improved speed/clarity for `step_spatialsign`


R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+      rlang::abort("cycle_unit column name must be a single character value.")
+
+    if (!is.na(starting_val) &
+        !is.numeric(starting_val) &
+        !inherits(starting_val, 'Date') &
+        !inherits(starting_val, 'POSIXt'))
+      rlang::abort("starting_val must be numeric, Date or POSIXt")
+
+
+    f_units <-
```
topepo, line 111: Use `rlang::arg_match0()`.

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+#'   'month_sidereal' (2551443 seconds), 'month_average' (2629744 seconds),
+#'   'week', day', 'hour', 'minute', 'second', 'sample'.
+#' @param starting_val either numeric, Date or POSIXt value that indicates the
+#'   reference point for the sin and cos curves.  This parameter may be
+#'   specified to increase control over the signal phase.  If starting_val is
+#'   not specified the default is the start of the data series, or the start of
+#'   the cycle (i.e. the beginning of the hour, day, week).
+#' @param objects Statistics are stored here once this step has
+#'  been trained by [prep.recipe()].
+#'
```
topepo, line 34: We'll need a Details section where you explain what this is doing and where it would be appropriate to use.

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+
+  if (length(frequency_1) > 0 & length(period_1) > 0) {
+    nms <- c(paste0('_f_', frequency_1), paste0('_p_', period_1))
+  } else if (length(frequency_1) > 0) {
+    nms <- paste0('_f_', frequency_1)
+  } else if (length(period_1) > 0) {
+    nms <- paste0('_p_', period_1)
+  }
+
+  col_names <- c(paste0(rep("sin", each = length(frequencies)), nms),
```
topepo, line 216: I don't like putting numeric data in column names (esp if it ends up being non-while numbers). Can you propose something different? Even an integer index would be good (with the tidy method giving use the key).

topepo: Either way, I would have the source data column name first in the new name.

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+}
+
+
+# may want a tunable step that can adjust the frequency/period
+#' #' @rdname tunable.step
+#' #' @export
+#' tunable.step_harmonic <- function(x, ...) {
+#'   tibble::tibble(
+#'     name = c("frequency", "period"),
+#'     call_info = list(
```
topepo, line 366: We can add functions to `dials` for this. I would give them placeholder names

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+#'       list(pkg = "dials", fun = ""),
```
topepo, line 367:
```suggestion
#'       list(pkg = "dials", fun = "harmonic_frequency"),
```

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+#'       list(pkg = "dials", fun = "")
```
topepo, line 368:
```suggestion
#'       list(pkg = "dials", fun = "harmonic_period")
```

R/harmonic.R
topepo: A reference would be good too.

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+  res
+}
+
+
+# may want a tunable step that can adjust the frequency/period
+#' #' @rdname tunable.step
+#' #' @export
+#' tunable.step_harmonic <- function(x, ...) {
+#'   tibble::tibble(
+#'     name = c("frequency", "period"),
```
topepo, line 365: `cycle_unit` too?

topepo (changes requested):

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+      cycle_unit = cycle_unit,
+      starting_val = starting_val,
+      objects = objects,
+      skip = skip,
+      id = id
+    )
+  }
+
+
+get_cycle_n <- function(x) {
```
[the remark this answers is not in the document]
jkennel: I removed this option to simplify inputs.  `cycle_size` provides a more flexible input.

R/harmonic.R
jkennel: No longer necessary.

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+           trained = FALSE,
+           frequency = NA_real_,
+           period = NA_real_,
+           cycle_unit = 'day',
+           starting_val = NA_real_,
+           objects = NULL,
+           skip = FALSE,
+           id = rand_id("harmonic")) {
+
+    if (!all(is.numeric(frequency)))
```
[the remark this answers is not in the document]
jkennel: Removed checks on frequency for tuning.

R/harmonic.R
jkennel: Added details with a couple of references and another example to provide a little more background.

R/harmonic.R
jkennel: I went with your integer suggestion. e.g. varname_sin_1, varname_sin_2, varname_cos_1, varname_cos_2,

R/harmonic.R
```diff
@@ -0,0 +1,374 @@
...
+step_harmonic <-
+  function(recipe,
+           ...,
+           role = NA,
+           trained = FALSE,
+           frequency = NA_real_,
+           period = NA_real_,
+           cycle_unit = 'day',
+           starting_val = NA_real_,
+           objects = NULL,
```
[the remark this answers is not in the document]
jkennel: Added.

R/harmonic.R
jkennel: I limit the input to frequency.  The user can do the period to frequency conversion outside the step.

R/harmonic.R
jkennel: Added.

R/harmonic.R
jkennel: I removed the input option for `period`, because it is essentially redundant with frequency.

R/harmonic.R
jkennel: They shouldn't tune over this or over `cycle_size`.  Now tuning over only `frequency` should be preferred all that is needed.

jkennel:
Sorry for the large numbers of changes.  I reduced the number of input arguments in an attempt to simplify the step and improve the experience.

R/harmonic.R
topepo: What is the range of that parameter?

topepo:
Sorry for the delay! Look great. Thanks

PR #737, opened 2021-06-28
updates to docs for recipe(), prep(), and bake()

### LINKED ISSUE #733: An alternative to `tidy` for inspecting recipe with intuitive name
Currently we need `tidy` function to to inspect recipe steps.
However, I think the name is not intuitive for several reasons

- The name does not contain `step`
    - This also make IDE to suggest the proper function when user guess the function contains `step` as its name.
- Too many `tidy` methods in the help

I'd suggest something like below.
I also add `.tidy` argument so that user can get raw result from the step when `.tidy = FALSE`.

```r
inspect_step <- function(x, number = NA, id = NA, ..., .tidy = TRUE) {
  if (.tidy) {
    return(tidy(x, number, id, ...)
  }
  if (!is.na(number)) {
    return(x$steps[[number]])
  }
  x$steps[[id]]
}
```

topepo:
closes #733

Related to #733 and tidymodels/TMwR#163, these changed highlight the use of `tidy()` on a recipe and also let users know that, for modeling, a workflow is the recommended approach.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -70,3 +70,6 @@ RdMacros:
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.1.9001
+Remotes:
+    tidymodels/parsnip
+    
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,6 +53,7 @@ Suggests:
     kernlab,
     knitr,
     modeldata,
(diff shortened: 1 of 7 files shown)


juliasilge:
I tried to make these a bit more consistent and clear, and I added another demo of and link to `tidy()` in the `recipe()` docs to further address the original issue.

juliasilge (approved):
These are some nice doc improvements 👍

PR #738, opened 2021-06-29
Fix in docs for signed argument in step_log

EdwinTh:


juliasilge:
Thank you so much @EdwinTh! 🙌

PR #749, opened 2021-07-14
Reduce repetition in documentation with `@inheritParams`

juliasilge:
This PR mostly focuses on using the documentation in `step_center()` and `step_pca()` for the other steps via `@inheritParams`, rather than repeating ourself a lot. It is not exhaustive (we still have more work to do in this area) but it focuses mostly on the `...` and `role`, with a few other lesser used params thrown in.

### DIFF
--- a/R/BoxCox.R
+++ b/R/BoxCox.R
@@ -5,11 +5,6 @@
 #'  transformation.
 #'
 #' @inheritParams step_center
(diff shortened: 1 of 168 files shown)


hfrick (approved):
Looks great! :purple_heart: :broom:

PR #753, opened 2021-07-16
Fix imputation steps on pkgdown site

### LINKED ISSUE #745: Not all imputation steps appears on pkgdown site
https://recipes.tidymodels.org/dev/reference/index.html#section-step-functions-imputation

Screen Shot 2021-07-08 at 11 28 42 AM
Screen Shot 2021-07-08 at 11 28 49 AM

juliasilge:
Closes #745 

When we changed the names of the imputation steps, we made the old versions `internal` but then we put them on the main `.Rd` files, disappearing the steps entirely. 🤪 This PR fixes this, the old versions not `internal` for the time being. They have to be documented as long as they are exported, and it makes the most sense to have them with the functions they are being replaced by.

This PR also escalates the old functions' deprecation to a warning.


hfrick (approved):
The title for the different steps is not particularly consistent (but that's the case for several sections) - do we want to change that here or in a different PR?

image

juliasilge:
Great point; I updated [these titles to use sentence case as outlined in the style guide](https://style.tidyverse.org/documentation.html#title-and-description). We can eventually harmonize the other ones.

PR #755, opened 2021-07-20
Doc touch up

hfrick:
We talk about a recipe as a sequence of operations (since they can be steps and checks) in the [template for the return value of checks](https://github.com/tidymodels/recipes/pull/749/commits/592c9f786858071508bf7753491ee26d390b3d64) so I've updated the template for the return value of step functions accordingly.


juliasilge (approved):
Great, nice 🙌

PR #787, opened 2021-09-09
Update doc titles to "preprocessing recipe"

### LINKED ISSUE #786: Use of term "data recipe"
The words "data recipe" are being used [3 times](https://github.com/tidymodels/recipes/search?q=%22data+recipe%22) in this package. Additionally, it is never used on https://tidymodels.org/ or https://juliasilge.com/

in the title of `prep()`

https://github.com/tidymodels/recipes/blob/433f2a8e589fb2875c1a07f6a6e49c7a251b16d1/R/recipe.R#L244

in the title of `bake()`

https://github.com/tidymodels/recipes/blob/433f2a8e589fb2875c1a07f6a6e49c7a251b16d1/R/recipe.R#L460

and as part of the `print.recipe()`

https://github.com/tidymodels/recipes/blob/433f2a8e589fb2875c1a07f6a6e49c7a251b16d1/R/recipe.R#L630-L632

I feel like it would be worth considering how often this phrase should be used. Either stop using it altogether to avoid potential confusion about what other types of recipes there are. Or expand the usage of the term across educational material.

juliasilge:
Closes #786 

I definitely agree with @EmilHvitfeldt that the phrase is a bit out of place and not consistent with how we now talk about recipes; I took a stab at changed wording here. I also updated to [sentence case for titles](https://style.tidyverse.org/documentation.html#title-and-description).

### DIFF
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -628,7 +628,7 @@ bake.recipe <- function(object, new_data, ..., composition = "tibble") {
 #' @author Max Kuhn
 #' @export
 print.recipe <- function(x, form_width = 30, ...) {
-  cat("Data Recipe\n\n")
+  cat("Recipe\n\n")
   cat("Inputs:\n\n")
   no_role <- is.na(x$var_info$role)
   if (any(!no_role)) {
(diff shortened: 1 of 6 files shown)


EmilHvitfeldt:
Looks good!
What about the print method

https://github.com/tidymodels/recipes/blob/433f2a8e589fb2875c1a07f6a6e49c7a251b16d1/R/recipe.R#L630-L632

PR #790, opened 2021-09-14
Penalized non-negative matrix factorization

topepo:
Much faster than the NMF package used by `step_nnmf()`. Results are about the same but the new step is over 500-fold faster. 

image

`step_nnmf()` is soft-deprecated. 

Unit tests will go into the `extratests` repo. 

Will probably wait to merge for the next release of `RcppML` by @zdebruine based on zdebruine/RcppML#5 (no hurry)

PR #796, opened 2021-09-14
Fix warnings seen while running `document()`

DavisVaughan:
These are related to trying to link to `tunable.step()` and `required_pkgs.step()` help pages, both of which don't exist


topepo:
We both met our frustration tolerances today. I have these changes in #790 😄

R/depth.R
```diff
@@ -210,13 +210,7 @@ tidy.step_depth <- function(x, ...) {
...
 
 
-
-#' S3 methods for tracking which additional packages are needed for steps.
-#'
-#' @param x A recipe step
-#' @return A character vector
-#' @rdname required_pkgs.step
-#' @keywords internal
+#' @rdname required_pkgs.recipe
```
DavisVaughan, line 213: Removed these duplicate docs from the step-depth method

R/recipe.R
```diff
@@ -779,7 +779,13 @@ utils::globalVariables(c("number"))
...
-#' @rdname required_pkgs
+#' S3 methods for tracking which additional packages are needed for steps.
+#'
+#' @param x A recipe or recipe step
+#' @param infra Should recipes itself be included in the result?
+#' @return A character vector
+#' @name required_pkgs.recipe
+#' @keywords internal
 #' @export
 required_pkgs.recipe <- function(x, infra = TRUE, ...) {
```
DavisVaughan, line 790: I went with naming the rd file `required_pkgs.recipe` rather than `required_pkgs.step` because I document it with the recipe method and I figured that made more sense

R/tunable.R
```diff
@@ -20,7 +20,6 @@
 #' The information about the default parameter object takes the form of a
 #' named list with an element for the function call and an optional element for
 #' the source of the function (e.g. the `dials` package).
-#' @export
```
DavisVaughan, line 23: Duplicate export tag here

R/recipe.R
```diff
@@ -792,13 +798,13 @@ required_pkgs.recipe <- function(x, infra = TRUE, ...) {
   res
 }
 
-#' @rdname required_pkgs
+#' @rdname required_pkgs.recipe
```
DavisVaughan, line 801: I think in your PR this still is `@rdname required_pkgs`, but I think that it has to link to the documentation for the specific recipe method to work right

DavisVaughan (commented):
Oh okay, I marked a few things I did in mine that you may want to do in yours. I'll let you decide how to proceed from here

topepo (approved):

PR #800, opened 2021-09-14
Fix formatting of bag imputation on factors

topepo:
There is a bug (reported to maintainer today) in `ipred::predict.classbag()` where the factor levels of the predictions may not be correct. This results in a reverse dependency breakage.

This PR

1. temporarily converts the outcome to a factor if a character column is to be imputed. The original data is untouched. 
1. casts the imputation predictions into a format consistent with raw data (factor, character, or numeric)
1. no longer casts the data to be a factor (to be consistent with `prep(rec, strings_as_factors = FALSE)`)

I also made changes to a $%&# Rmd file that is always changing because of `cli::rule()` widths and random step id's.


juliasilge:
In workflows, we [handle the printing width like this](https://github.com/tidymodels/workflows/blob/84ae7afc39ae0ad30ce2dc254639e0da01c530bb/man/rmd/indicators.Rmd#L4):

```
options(cli.width = 70, width = 70, cli.unicode = FALSE)
```

I think this is a really good think to make sure we have so this is more consistent. 👍 

I agree that the random IDs are frustrating, but if you look at [how this documentation ends up](https://recipes.tidymodels.org/reference/recipe.html), we don't ever explain about what `id` does. It seems like a bad idea to use it in the documentation as it exists now.

We could add more explanation, like [what we have here](https://www.tmwr.org/recipes.html#tidy-a-recipe), but this page is already getting really long-winded and detailed.

- Do we prefer to add this discussion here so we can use `id`? (I don't think I would want to add it on its own merits TBH.)
- Do we want to get used to not staging this chunk? This is what I typically do when I run `devtools::document()`; I either don't stage this whole file or if I need something from this file, I don't stage that chunk.

DavisVaughan:
I fixed the random step id's here by setting the seed. We shouldn't have to worry about that anymore
https://github.com/tidymodels/recipes/pull/794

DavisVaughan:
My guess on the ipred bug without running any code is that they probably need to set `levels = classes` here:
https://github.com/cran/ipred/blob/9c7b29aea310803f5d55aa229ee9519847a192e3/R/predict.bagging.R#L54

i.e.

```r
RET <- factor(classes[apply(vote, 1, uwhich.max)], levels = classes)
```

I imagine you have a case where `classes[apply(vote, 1, uwhich.max)]` isn't resulting in one of each level, so `factor()` doesn't know about the missing levels

R/impute_bag.R
```diff
@@ -262,8 +266,8 @@ bake.step_impute_bag <- function(object, new_data, ...) {
         rlang::warn("All predictors are missing; cannot impute")
       } else {
         pred_vals <- predict(object$models[[imp_var]], pred_data)
-        pred_vals <- cast(pred_vals, new_data[[imp_var]])
-        new_data[[imp_var]] <- vec_cast(new_data[[imp_var]], pred_vals)
+        # For an ipred bug reported on 2021-09-14:
+        pred_vals <- cast(pred_vals, object$models[[imp_var]]$y)
```
DavisVaughan, line 270: Is this always a factor output?

If so, this could probably be as simple as:

```r
pred_vals <- factor(pred_vals, levels = levels(object$models[[imp_var]]$y))
```

That would get more at the heart of the bug, I think

Or you could wrap that in `if (is.factor(pred_vals)) { <code> }` if `pred_vals` could be numeric

R/impute_bag.R
```diff
@@ -180,6 +180,10 @@ step_impute_bag_new <-
 bag_wrap <- function(vars, dat, opt, seed_val) {
   seed_val <- seed_val[1]
   dat <- as.data.frame(dat[, c(vars$y, vars$x)])
+  if (is.character(dat[[vars$y]])) {
+    dat[[vars$y]] <- factor(dat[[vars$y]])
+  }
```
DavisVaughan, line 185: Is this because `ipred::ipredbagg()` doesn't have a character method? So you've been getting this message?

```r
> ipred:::ipredbagg.default
function (y, ...) 
{
    stop(paste("Do not know how to handle objects of class", 
        class(y)))
}
```

R/impute_bag.R
topepo: No, it could be numeric or factor. Is there are reason to get away from `cast()`?

R/impute_bag.R
topepo: Yes.

topepo:
Torsten fixed the `ipred` bug within like 15 and the new version is [on CRAN](https://cran.r-project.org/web/packages/ipred/index.html)

R/impute_bag.R
DavisVaughan: I guess it looks like it does the same thing. I was just thinking about the most minimal change that targets this bug, so we can safely remove it later on once it's fixed. It doesn't really matter

PR #801, opened 2021-09-15
Update to testthat 3e

DavisVaughan:
This PR updates us to testthat 3e. The main changes are:

- Removal of `expect_equivalent()`
- Addition of `expect_snapshot()` in some places
- Silencing some chatty functions now that messages and multiple warnings are shown in testthat
   - In particular, multiple dimRed functions emit messages with no option to silence them. So I silence them with `suppressMessages()`
- Fix a bug with `dummy_names()` and zero length input

@topepo I don't think this needs a super in depth review. But if you could take a quick glance at this that would be great.


topepo:
It looks great. Thanks

PR #802, opened 2021-09-17
Deprecation clean-up

### LINKED ISSUE #762: undo deprecation of `step_kpca()`
it was deprecated in version 0.1.7

hfrick:
This PR does a few things with regard to deprecation and sneaks in a bonus clean-up of the pkgdown site:

* undoes the deprecation of `step_kpca()`; closes #762
* escalates the deprecation of the `preserve` argument to `step_pls()` and `step_dummy()`
* adds lifecycle-style deprecation for the `options` argument to `step_nzv()` so it won't get overlooked next time 
* tidies up the pkgdown site so that the newly exported `recipes_eval_select()` shows up in the section for internal tools rather than at the top and removes the quadruplication of the section on selection via role

The imputation steps and the steps for up/downsampling were already in the correct stage of the deprecation process.

### DIFF
--- a/R/nzv.R
+++ b/R/nzv.R
@@ -82,14 +82,12 @@ step_nzv <-
     if (!isTRUE(all.equal(exp_list, options))) {
       freq_cut <- options$freq_cut
       unique_cut <- options$unique_cut
-      message(
-        paste(
-          "The argument `options` is deprecated in favor of `freq_cut`",
-          "and `unique_cut`. options` will be removed in next version."
-        )
+      lifecycle::deprecate_stop(
+        "0.1.7",
+        "step_nzv(options)",
+        details = "Please use the arguments `freq_cut` and `unique_cut` instead."
       )
     }
-
 
     add_step(
       recipe,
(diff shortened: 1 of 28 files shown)


R/nzv.R
```diff
@@ -82,15 +82,13 @@ step_nzv <-
...
-      message(
-        paste(
-          "The argument `options` is deprecated in favor of `freq_cut`",
-          "and `unique_cut`. options` will be removed in next version."
-        )
+      lifecycle::deprecate_stop(
+        "0.1.7",
+        "step_nzv(options)",
+        details = "Please use the arguments `freq_cut` and `unique_cut` instead."
       )
```
hfrick, line 89: this has been deprecated for quite a while but since it only ever sent a message, but no warning or error, I didn't remove it right away but have it throw an error now

juliasilge (approved):
This looks great 🙌 

If we are keeping all three kPCA steps, that puts us at "copying three times" and I say let's put the info in a roxygen template. The template is technically incorrect right now for `step_kpca()` (no `keep_original_cols = FALSE`) but I will work on that immediately after this is merged in.

I'll let you merge @hfrick after you look at my work here.

hfrick:
Oh, good catch! Thanks for adding that. :raised_hands: 

I've added a single "the" to the sentence and then it cascaded across the standard text width so it looks like a lot more changes than that :eyes: 

I think this is ready to be merged. I'll do that once the checks are finished

PR #805, opened 2021-09-17
export `tidy()` methods for `check_new_values()`, `check_class()`, and `step_nnmf()`

### LINKED ISSUE #731: tidy.check_new_values is unexported
This is inconsistent with other `check_*` functions and appears to be mistakenly omitted

https://github.com/tidymodels/recipes/blob/2f73e8c400351f906494ede06c0b0961b358475c/R/newvalues.R#L161-L169

### LINKED ISSUE #809: Export `tidy.check_class()` and `tidy.step_nnmf()`
The tidy methods for the other checks and steps are exported so this seems like an oversight.

hfrick:
closes #731, assuming it's only an oversight that the tidy method for this check isn't exported when the other ones are

 closes #809


R/newvalues.R
```diff
@@ -143,6 +143,9 @@ print.check_new_values <-
     invisible(x)
   }
 
+#' @rdname tidy.recipe
+#' @param x A `check_new_values` object.
```
DavisVaughan, line 147: I don't think this `@param` tag is required

Unfortunately I have discovered that we have used this param tag in all of the tidy methods, but since they are all documented on the same page, only the _last_ one that roxygen2 sees is actually used as the parameter documentation for `x`, which I think is the one for `step_zv()` since that is last alphabetically.

I think we can really just remove them all (in a separate PR) if we wanted to, and only keep a single source of documentation in `tidy.recipe`. Nevertheless, I'd remove it from here

DavisVaughan (approved):
Yea, looks like an oversight! Thanks

R/newvalues.R
hfrick: sounds good :+1:

hfrick:
While cleaning up the `@param` tag across tidy methods for various steps and checks I found two more: `tidy.check_class()` and `tidy.step_nnmf()` are also not exported. Would it be better to add them here to this PR or make a new one, @DavisVaughan ?

DavisVaughan:
Here is probably fine!

hfrick:
added them here! closes #809

PR #816, opened 2021-09-22
remove curly braces from @family tags

topepo:
We were using syntax such as `@family {variable filter steps}`. These led to R CMD check issues: 

```
Found the following \keyword or \concept entries with Rd markup:
  File 'check_class.Rd':
    \concept{{checks}}
  File 'check_cols.Rd':
    \concept{{checks}}
  File 'check_missing.Rd':
    \concept{{checks}}
```


NAMESPACE
```diff
@@ -223,7 +223,6 @@ S3method(required_pkgs,step_pls)
 S3method(required_pkgs,step_window)
 S3method(summary,recipe)
 S3method(tidy,check)
-S3method(tidy,check_class)
```
juliasilge, line 226: I think we do still want to export this, right?

juliasilge: Related to #805

NAMESPACE
```diff
@@ -282,7 +281,6 @@ S3method(tidy,step_modeimpute)
 S3method(tidy,step_mutate)
 S3method(tidy,step_mutate_at)
 S3method(tidy,step_naomit)
-S3method(tidy,step_nnmf)
```
juliasilge, line 285: Same as above ⬆️

PR #830, opened 2021-10-08
move methods for `tune_args()` from tune to here

hfrick:
This PR is part of moving methods for `tune_args()` from tune to the package which holds the corresponding object class: 

* the methods for `recipe`, `step`, and `check` are moved to recipes (aka this PR here)
* the method for `model_spec` is moved to parsnip 
* the method for `workflow` is moved to workflows 

In all cases, the methods are registered conditionally on the version of tune from which the methods have been removed. The tune PR is FIXME

The tests require multiple packages so they are all in extratests:  https://github.com/tidymodels/extratests/pull/30

The PR on extratests should be merged first, then the ones on recipes and parsnip, then workflows, then tune.

This is part of the wider aim of introducing `extract_parameter_dials()` and `extract_parameter_set_dials()` and the next steps are:
* moving the methods for `tunable()` similar to this PR set
* moving the methods for `parameters()` similar to this PR set 
* introducing `extract_parameter_set_dials()` 
* introducing `extract_parameter_dials()` 
* updating documentation and learning materials


R/tune_args.R
```diff
@@ -0,0 +1,190 @@
...
+  )
+}
+
+
+# @export - lazily and conditionally registered in .onLoad()
+tune_args_check <- tune_args_step
+
+
+# helpers for tune_args() methods -----------------------------------------
+# these are copied from tune
```
hfrick, line 60: These are used in parsnip as well so I plan to look into exporting them in tune and using the exported versions here, instead of the copy.

DavisVaughan: Hmm I am not sure you can export these from tune and use them here.

recipes/parsnip don't rely on tune in any way, and the `tune_args()` methods for model-spec and recipe should be able to run without calling anything from tune.

So these helpers might have to just be duplicated in recipes and parsnip, which is ok

R/reexports.R
```diff
@@ -11,6 +11,10 @@ generics::required_pkgs
 #' @export
 generics::tunable
 
+#' @importFrom generics tune_args
+#' @export
+generics::tune_args
```
DavisVaughan, line 16: Same general comment about the fact that I wouldn't export `tune_args()`. But we already export `tunable()` here so maybe it is worth it to be consistent for the recipes case...

R/zzz.R
```diff
@@ -0,0 +1,23 @@
+# nocov start
+
+.onLoad <- function(libname, pkgname) {
+
+  # Can't use `rlang::is_installed()` at all, as that doesn't work when
+  # called from `.onLoad()` for some reason. Instead, rely on `packageVersion()`
+  # erroring when the package isn't installed.
+  has_at_least_version <- function(pkg, version) {
```
DavisVaughan, line 8: Needs same changes as workflows

R/tune_args.R
hfrick: The helpers are not used in tune itself any more. So I'd remove them from tune and make the comments in recipes and parsnip point at each other so that we know that the same code is in both packages. Or are the comments overkill?

DESCRIPTION
```diff
@@ -72,3 +72,5 @@ Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.2
 Config/testthat/edition: 3
+Remotes: 
+    r-lib/generics
```
DavisVaughan, line 76: We are going to wait until generics hits cran to merge this, and we will update to use CRAN generics once that happens

DavisVaughan (approved):

PR #835, opened 2021-10-13
case weights

topepo:
I'll implement the steps sequentially starting off with a draft PR to get review on the changes to `step_center()`.


R/selections.R
```diff
@@ -172,7 +172,12 @@ recipes_eval_select <- function(quos, data, info, ..., allow_rename = FALSE) {
 
   local_current_info(nested_info)
 
-  expr <- expr(c(!!!quos))
+  if (is_quosures(quos)) {
```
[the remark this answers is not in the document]
DavisVaughan: You have two options:

You could use `enquos(case_weights)` (i.e. the plural version that gives you a list of quosures) so you don't have to special case this in `recipes_eval_select()`.

Ideally you'd use `enquo(case_weights)` like you have, but `recipes_eval_select()` expects a list of quosures so we need to give it one somehow

---

Alternatively in `get_case_weights()` you can probably wrap `selection` in a list like:

```r
wt_col <- unname(recipes_eval_select(list(selection), .data, info))
```

That may be better since you can add a comment here about why you have to wrap it in a list (because `recipes_eval_select()` wants one), but otherwise you can continue to correctly use `enquo(case_weights)`

If all of the case weight supported steps go through `get_case_weights()`, then that sounds good to me

tests/testthat/test-case-weight-functions.R
```diff
@@ -0,0 +1,84 @@
+library(testthat)
+source(test_path("helper-case-weights.R"))
```
juliasilge, line 2: Anything called "helper" is [automatically run before tests and loaded via `load_all()`](https://blog.r-hub.io/2020/11/18/testthat-utility-belt/) so we don't need to source this. We might prefer to change the file to "setup-case-weights.R" so those objects aren't in memory when we run `load_all()`

juliasilge (commented):
I worked through this for `step_center()` and I really like where this landed. IMO this is very ergonomic and consistent with how recipes' overall design.

R/center.R
```diff
@@ -7,6 +7,10 @@
 #'  sequence of operations for this recipe.
 #' @param ... One or more selector functions to choose variables
 #'  for this step. See [selections()] for more details.
+#' @param case_weights A selector that specifies a single numeric column that
```
topepo, line 10: Since we have the case weight classes and roles working, we won't use arguments for the case weights. They will automatically get picked up by the steps and used (or not) depending on the type of weight.

R/case_weights.R
```diff
@@ -0,0 +1,202 @@
...
+#' For `averages()` and `variances()`, missing values in the data (*not* the
+#' case weights) only affect the calculations for those rows. For
+#' `correlations()`, the correlation matrix computation first removes rows
+#' with any missing values (equal to the "complete.obs" strategy in
+#' [stats::cor()]).
+#'
+#' `is_unsupervised_weights()`
+#' @export
+#' @name case-weight-helpers
+get_case_weights <- function(selection, info, .data) {
```
topepo, line 31: This can be simplified. The recipe already checks the type and number of case weights. Since the weights will not be arguments to the steps, there is not need of quosures or tidyselect.

R/center.R
EmilHvitfeldt: Okay. It would make it cleaner to disallow the user from turning off case weights on a step by step basis.

EmilHvitfeldt:
I added preliminary printing support for case weights. Applied in `step_center()` and `step_dummy_extract()`. `step_center()` can be used with both importance and frequency weights, `step_dummy_extract()` is only supported for frequency weights.

``` r
library(recipes)

mtcars1 <- mtcars %>%
  tibble::rownames_to_column() %>%
  mutate(wt = importance_weights(wt))

recipe(mpg ~ ., data = mtcars1) %>%
  step_inverse(drat) %>%
  step_center(all_numeric_predictors()) %>%
  step_dummy_extract(rowname, sep = " ") %>%
  prep()
#> Recipe
#> 
#> Inputs:
#> 
#>          role #variables
#>  case_weights          1
#>       outcome          1
#>     predictor         10
#> 
#> Training data contained 32 data points and no missing data.
#> 
#> Operations:
#> 
#> Inverse transformation on drat [trained]
#> [Weighted] Centering for cyl, disp, hp, drat, qsec, vs, am, gear, carb [trained]
#> [Not weighted] Extract patterns from rowname [trained]
```

PR #861, opened 2021-12-07
Add step_filter_missing()

### LINKED ISSUE #270: add a step to filter columns using missingness
If there are more than _X_% missing, remove the column

EmilHvitfeldt:
This PR aims to close #270.

The step only has one parameters `threshold`, defaulted to `0.1`. A good side-effect is that setting `threshold = 0` removed all columns with missing values. Should work on any type of columns.


R/filter_missing.R
```diff
@@ -0,0 +1,150 @@
+#' Missing Value Column Filter
+#'
+#' `step_filter_missing` creates a *specification* of a recipe
+#'  step that will potentially remove variables that have too many missing
+#'  values.
+#'
+#' @inheritParams step_center
+#' @param threshold A value for the threshold of missing values in column. The
+#'  step will try to remove the columns where the proportion of missing values
```
juliasilge, line 9:
```suggestion
#'  step will remove the columns where the proportion of missing values
```

R/filter_missing.R
```diff
@@ -0,0 +1,150 @@
...
+#'  exceeds the threshold.
+#' @param removals A character string that contains the names of
+#'  columns that should be removed. These values are not determined
+#'  until [prep.recipe()] is called.
+#' @template step-return
+#' @template filter-steps
+#' @family variable filter steps
+#' @export
+#'
+#' @details This step attempts to remove variables if the proportion of missing
```
juliasilge, line 19:
```suggestion
#' @details This step will remove variables if the proportion of missing
```

R/filter_missing.R
```diff
@@ -0,0 +1,150 @@
...
+#' @export
+tidy.step_filter_missing <- tidy_filter
+
+#' @rdname tunable.recipe
+#' @export
+tunable.step_filter_missing <- function(x, ...) {
+  tibble::tibble(
+    name = "threshold",
+    call_info = list(
+      list(pkg = "dials", fun = "threshold")
```
topepo, line 143: We should add a range here (maybe `c(0.05, 1.00)`?)

_pkgdown.yml
```diff
@@ -109,6 +109,7 @@ reference:
     - step_rm
     - step_select
     - step_zv
+    - step_filter_missing
```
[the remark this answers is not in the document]
EmilHvitfeldt: Thank you!

PR #863, opened 2021-12-07
Add group argument to step_zv()

### LINKED ISSUE #711: conditional zero-variance filter
`discrim` now has a quadratic discriminant model and that requires 2+ unique values within each class. A new `step_zv_group()` could do the same filter but look for zero-variance within 1 or more columns that define groups.

EmilHvitfeldt:
This PR aims to close #711.

`group` can be one or more character strings or a call to `dplyr::vars()`.
`group` Is currently not restricted to refer to any types. Hence the examples before refer to two integer variables to define the groups.

Should I expand this to work for `step_nzv()` as well since the steps are so similar?

``` r
library(recipes)

recipe(~ ., data = mtcars) %>%
  step_zv(all_predictors(), group = "am") %>%
  prep() %>%
  tidy(1)
#> # A tibble: 0 × 2
#> # … with 2 variables: terms <chr>, id <chr>

recipe(~ ., data = mtcars) %>%
  step_zv(all_predictors(), group = c("am", "vs")) %>%
  prep() %>%
  tidy(1)
#> # A tibble: 2 × 2
#>   terms id      
#>   <chr> <chr>   
#> 1 cyl   zv_zSLT3
#> 2 gear  zv_zSLT3

recipe(~ ., data = mtcars) %>%
  step_zv(all_predictors(), group = vars(am, vs)) %>%
  prep() %>%
  tidy(1)
#> # A tibble: 2 × 2
#>   terms id      
#>   <chr> <chr>   
#> 1 cyl   zv_bnxSp
#> 2 gear  zv_bnxSp
```

<sup>Created on 2021-12-07 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/zv.R
+++ b/R/zv.R
@@ -7,8 +7,11 @@
 #' @param removals A character string that contains the names of
 #'  columns that should be removed. These values are not determined
 #'  until [prep.recipe()] is called.
-#' @param group An optional character string or call to [dplyr::vars()] that can be used to
-#'  specify a group(s) within which to identify variables that contain only a single value.
+#' @param group An optional character string or call to [dplyr::vars()]
+#'  that can be used to specify a group(s) within which to identify
+#'  variables that contain only a single value. If the grouping variables
+#'  are contained in terms selector, they will not be considered for
+#'  removal.
 #' @template step-return
 #' @template filter-steps
 #' @details When you [`tidy()`] this step, a tibble with column `terms` (the
--- a/R/zv.R
+++ b/R/zv.R
@@ -7,8 +7,8 @@
 #' @param removals A character string that contains the names of
 #'  columns that should be removed. These values are not determined
 #'  until [prep.recipe()] is called.
-#' @param group A character string or call to dplyr::vars() can be used to
-#'  specify the group(s) variable.
+#' @param group An optional character string or call to [dplyr::vars()] that can be used to
+#'  specify a group(s) within which to identify variables that contain only a single value.
 #' @template step-return
 #' @template filter-steps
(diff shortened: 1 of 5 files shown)


R/zv.R
```diff
@@ -7,6 +7,8 @@
 #' @param removals A character string that contains the names of
 #'  columns that should be removed. These values are not determined
 #'  until [prep.recipe()] is called.
+#' @param group A character string or call to dplyr::vars() can be used to
+#'  specify the group(s) variable.
```
juliasilge, line 11:
```suggestion
#' @param group An optional character string or call to [dplyr::vars()] that can be used to
#'  specify a group(s) within which to identify variables that contain only a single value.
```

R/zv.R
```diff
@@ -7,6 +7,8 @@
 #' @param removals A character string that contains the names of
 #'  columns that should be removed. These values are not determined
 #'  until [prep.recipe()] is called.
+#' @param group An optional character string or call to [dplyr::vars()] that can be used to
+#'  specify a group(s) within which to identify variables that contain only a single value.
```
topepo, line 11: Please make a note that any variables used here will be removed from `terms`.

EmilHvitfeldt: That is currently not happening. Are you saying that 

```
recipe(~ ., data = mtcars) %>%
  step_zv(all_predictors(), group = "am")
```

Should remove `am` as well as the zero variance columns?

topepo: No, I mean that if the terms selector is something like `all_predictors()` and their grouping variable is within `all_predictors()`, the grouping variable will never be removed.

Issue #453, opened 2020-01-23
Use of `step_nzv(eveything())` in combination with `step_intercept()`


EdwinTh:
I routinely use `step_nzv()` if factors are used in the recipe. This had me puzzled for a few minutes.:
```r
recipe(mtcars) %>% step_intercept() %>% whole bunch of other steps %>% 
step_nzv(everything()) %>% prep() %>% bake()
```
did not have an intercept. 

This is of course because the intercept is invariant. Technically it is completely correct, but it is of course never intended. You could leave this as the responsibility of the user, but I think you should at least throw a warning if `step_nzv()` removes the intercept that was added in that very recipe. Even better imo, is not removing the intercept, even when `everything()` is used. I cannot think of scenario you want to have it removed again, after adding it in the same recipe.

topepo:
We could check to see if there is a column called "intercept" except that 1) it might be called something else (since that is parameterized) and 2) there might be some other context where a user has a legit column called "intercept". Believe it or not, I can think of a use case for the latter.

EmilHvitfeldt:
We could use `recipes:::detect_step(recipe, "intercept")` inside `step_nzv()` to catch if `step_zv/nzv()` comes after `step_intercept()` and emit an warning.

topepo:
I think that we will just close this. The simple solution is to add it as the last step or to deselect it in subsequent steps. The proposed solutions raise the code complexity without a whole lot of benefit.

PR #868, opened 2021-12-13
make tidy.recipe work for zero step recipes

### LINKED ISSUE #867: tidy.recipe() throws error for zero-step recipes
`tidy.recipe()` should return a zero row tibble for empty recipes instead of throwing an error. This is making `step_detect()` error when used on empty recipes.

``` r
library(recipes)

recipe(~., data = mtcars) %>%
  tidy()
#> # A tibble: 1 × 6
#>   number operation type   trained skip  id          
#>    <int> <chr>     <chr>  <lgl>   <lgl> <chr>       
#> 1      1 step      center FALSE   FALSE center_uXNNk

recipe(~., data = mtcars) %>%
  tidy()
#> Error: No steps in recipe.
```

EmilHvitfeldt:
This PR aims to close #867.

Very simple change. I need it to fix `detect_step()` for https://github.com/tidymodels/recipes/issues/453


R/tidy.R
```diff
@@ -55,8 +55,16 @@ tidy.recipe <- function(x, number = NA, id = NA, ...) {
...
+  if (num_oper == 0) {
+    return(
+      tibble(number = integer(),
+             operation = character(),
+             type = character(),
+             trained = logical(),
+             skip = logical(),
+             id = character())
+    )
+  }
```
DavisVaughan, line 67: What happens if we just remove this branch entirely? It looks like our usage of `vapply()` should set us up pretty well for the 0-step case to "just work" without special handling? Maybe one or two calls to some base R function might need to be tweaked.

EmilHvitfeldt: Good eye! Appears to work without further modification

PR #882, opened 2022-01-07
Add step_percentile

### LINKED ISSUE #765: Request `step_percentile()` , non-obvious opportunity for data leak with `step_mutate()`

EmilHvitfeldt:
This PR aims to close #765 by adding `step_percentile()`. The code is pretty much a copy from https://www.tidymodels.org/learn/develop/recipes/ with the changes localized in this commit https://github.com/tidymodels/recipes/commit/ba42ad5477d3d128b5ea1ea6c546787293414a59


juliasilge:
If we add this as a new step, then we'll need to rewrite the tidymodels.org article with some new example transformation; it doesn't make much sense IMO to have the article for an existing step that people can use already.

EmilHvitfeldt:
how are we going to handle this long term? If the step we showcase in tidymodels.org is too useful will we end up getting requests for it, but it also needs to be useful enough to showcase. 

Could the example be a step which is a subset of an existing step? e.i. "this is how you can create a step that does a 2-degree polynomial expansion. If you need a general polynomial step you can use `step_poly()`". (I'm not saying `step_poly()` would be the right step to use here)

tests/testthat/test_percentile.R
```diff
@@ -0,0 +1,122 @@
...
+    biomass_tr_baked$carbon
+  )
+  expect_equal(
+    approx(carbon_quantiles, y = 0:100, xout = biomass_te$carbon)$y/100,
+    biomass_te_baked$carbon
+  )
+  sulfur_values <- as.numeric(gsub("%$", "", names(sulfur_quantiles)))
+  expect_equal(
+    approx(sulfur_quantiles, y = sulfur_values, xout = biomass_tr$sulfur)$y/100,
+    biomass_tr_baked$sulfur
```
topepo, line 42: Can you add a test case where there are fewer unique values in the column being transformed than percentiles requested? Maybe make a new variable that is something like `round(biomass$carbon, -1)`.

EmilHvitfeldt: Added in https://github.com/tidymodels/recipes/pull/882/commits/e204adbe4bbcc688d7c78f6001473d6f57647842

PR #900, opened 2022-02-09
Add tests for recipes_eval_select

EmilHvitfeldt:
This PR adds a suite of tests for `recipes_eval_select()`. These are slightly modified from the `terms_select()` tests. They additionally adds tests for `for all_numeric_predictors()` and `all_nominal_predictors()` which would close https://github.com/tidymodels/recipes/issues/841.

### DIFF
--- a/tests/testthat/_snaps/selections.md
+++ b/tests/testthat/_snaps/selections.md
@@ -1,0 +1,25 @@
+# simple name selections
+
+    Code
+      recipes_eval_select(quos = quos(log(date)), data = okc, info = info1)
+    Error <rlang_error>
+      non-numeric argument to mathematical function
+      Caused by error in `log()`:
+      ! non-numeric argument to mathematical function
+
+---
+
+    Code
+      recipes_eval_select(quos = quos(I(date:age)), data = okc, info = info1)
+    Error <rlang_error>
+      object 'age' not found
+      Caused by error in `unique.default()`:
+      ! object 'age' not found
+
+---
+
+    Code
+      recipes_eval_select(data = okc, info = info1)
+    Error <simpleError>
+      argument "quos" is missing, with no default
+
--- a/tests/testthat/_snaps/selections.md
+++ b/tests/testthat/_snaps/selections.md
@@ -10,15 +10,6 @@
 ---
 
     Code
-      recipes_eval_select(quos = quos(I(date:age)), data = okc, info = info1)
-    Error <rlang_error>
-      object 'age' not found
-      Caused by error in `unique.default()`:
-      ! object 'age' not found
-
----
-
-    Code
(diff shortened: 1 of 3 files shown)


tests/testthat/test-selections.R
```diff
@@ -0,0 +1,213 @@
+library(testthat)
+library(recipes)
+library(tibble)
+library(tidyselect)
+library(rlang)
```
DavisVaughan, line 5: I feel quite strongly that tibble, tidyselect, and rlang should not be `library()`ied here. That makes them available to all tests, which goes against the idea that this test file should be self contained. Just prefix their calls with `rlang::` etc, or if we have done `@import rlang` then you don't need this at all.

I personally don't think testthat or recipes should be libraried either, as `load_all()` and `check()` always loads them for you. But if Max needs them, then that is fine

tests/testthat/test-selections.R
```diff
@@ -0,0 +1,213 @@
...
+
+library(modeldata)
```
DavisVaughan, line 7: Same here if we can avoid this and still get the data out

tests/testthat/test-selections.R
```diff
@@ -0,0 +1,213 @@
...
+    "non-numeric argument to mathematical function"
+  )
+  expect_error(
+    recipes_eval_select(quos = quos(I(date:age)), data = okc, info = info1),
+    "object 'age' not found"
+  )
+  expect_error(
+    recipes_eval_select(data = okc, info = info1),
+    'argument "quos" is missing, with no default'
+  )
```
DavisVaughan, line 111: You should try using `expect_snapshot(error = TRUE, {expression})` here. Expecting an exact match against an error that you don't own is fragile, and `expect_snapshot()` won't fail on CRAN if this error ever changes (doubtful that it would, but still)

DavisVaughan (changes requested):

EmilHvitfeldt:
All requested changes were incorporated!

tests/testthat/_snaps/selections.md
```diff
@@ -0,0 +1,25 @@
+# simple name selections
+
+    Code
+      recipes_eval_select(quos = quos(log(date)), data = okc, info = info1)
+    Error <rlang_error>
+      non-numeric argument to mathematical function
+      Caused by error in `log()`:
+      ! non-numeric argument to mathematical function
```
DavisVaughan, line 8: this double showing of the error is strange, i dont think it is anything youve done though

DavisVaughan (approved):

tests/testthat/_snaps/selections.md
EmilHvitfeldt: Perfect, thank you!

PR #903, opened 2022-02-14
add skip for suggested package

MichaelChirico:


juliasilge:
`step_kpca()` uses kernlab; can you tell us more about what problem you ran into?

juliasilge:
Hmmmm, I just did `remove.packages("dimRed")` and then ran `devtools::test()`, and I did not have any problems. I don't believe we need dimRed at all for this test. Maybe you can be more specific in what you ran into?

MichaelChirico:
I see... sorry about that. I think I know what happened. but it's an internal issue of ours... closing for now.

PR #913, opened 2022-02-23
Add a version requirement for modeldata

MichaelChirico:
Since `tate_text` is new to that version & used several times in tests/examples


juliasilge (approved):
Thank you so much @MichaelChirico! 🚀

Issue #920, opened 2022-02-25
Add warning if step_normalize introduces NaN's in column


FieteO:
## Background
For columns with zero variance, `step_normalize()` replaces the whole column with `NaN`'s. I had a very hard time debugging this when trying to feed the data to a model and got an `Error: Missing data in columns:`.

## Reproducible example
``` R
x1 <- sample(c(NaN,5,10),300,replace=TRUE)
x2 <- sample(1,300,replace=TRUE)
x3 <- sample(c(0,1),300,replace=TRUE)
df1 <- data.frame(x1,x2,x3)

dfsplits <- initial_split(df1, strata = x3)
df_train <- training(dfsplits)
df_test  <- testing(dfsplits)

df_recipe <-
  recipe(x1 ~ ., data = df_train) %>%
  step_normalize(all_numeric_predictors())

df_recipe %>%
  prep(df_train) %>%
  bake(df_test)

# A tibble: 76 x 3
      x2     x3    x1
   <dbl>  <dbl> <dbl>
 1   NaN  0.946     5
 2   NaN -1.05    NaN
 3   NaN  0.946     5
 4   NaN -1.05      5
 5   NaN -1.05     10
 6   NaN  0.946    10
 7   NaN  0.946    10
 8   NaN  0.946    10
 9   NaN  0.946   NaN
10   NaN -1.05     10
# ... with 66 more rows
```

``` R
df_model <- rand_forest(mtry=50,trees=150) %>%
            set_engine('ranger',importance='impurity') %>%
            set_mode('regression')

df_workflow <- 
  workflow() %>% 
  add_model(df_model) %>% 
  add_recipe(df_recipe)

df_workflow %>% 
  fit(data = df_train)
# Error: Missing data in columns: x2.
```
I now know that this can be mitigated for this particular case with `step_zv()` to remove the zero-variance column, but it would be great if there had been any indication that `step_normalize()` introduced the `NaN` values.

juliasilge:
After we compute the standard deviations:

https://github.com/tidymodels/recipes/blob/d3c9240c63db800a1b8ed2255227711f3698b1d5/R/normalize.R#L109

we could check to see if any are ~zero, like `sds < .Machine$double.eps`, and throw an error saying which one is. (We could suggest using `step_zv()` too.) 

We would also want to do this for `step_scale()`.

EmilHvitfeldt:
This issue was closed by https://github.com/tidymodels/recipes/pull/927 thanks to @brunocarlin

PR #927, opened 2022-03-10
add errors for zv normalize and scale

brunocarlin:
This pr tries to solve the feature request on #920.

I have chosen to implement abort instead of warn, but we can also talk about the messaging when the error occurs.


R/scale.R
```diff
@@ -110,6 +110,20 @@ prep.step_scale <- function(x, training, info = NULL, ...) {
 
   sds <- sds * x$factor
 
+  which_sd <- which(sds ==0)
```
juliasilge, line 113: It doesn't work great in all situations to compare to zero with a numeric value like this because of how floating point numbers are stored. Instead, you can see if it is less than the machine tolerance with `sds < .Machine$double.eps` or `all.equal()` or something like that.

R/scale.R
```diff
@@ -110,6 +110,20 @@ prep.step_scale <- function(x, training, info = NULL, ...) {
...
+  if (length(which_sd) > 0) {
+    with_na <- names(which_sd)
+    with_na_str <- paste(paste0("`", with_na, "`"), collapse = ", ")
```
juliasilge, line 116: recipes already imports [glue](https://glue.tidyverse.org/) so let's use that.

R/scale.R
```diff
@@ -110,6 +110,20 @@ prep.step_scale <- function(x, training, info = NULL, ...) {
...
+    rlang::abort(c(
+      "Some columns have Zero variance",
+      "i" = 'Consider running step_zv to remove those columns.',
```
juliasilge, line 119:
```suggestion
      "Some columns have zero variance so normalization is not defined:",
      glue_cols,
      "i" = 'Consider `step_zv()` to remove those columns.',
```
Put the column names before the instruction about `step_zv()`.

R/scale.R
```diff
@@ -110,6 +110,20 @@ prep.step_scale <- function(x, training, info = NULL, ...) {
 
   sds <- sds * x$factor
 
+  which_sd <- which(sds ==0)
+  if (length(which_sd) > 0) {
+    with_na <- names(which_sd)
+    with_na_str <- paste(paste0("`", with_na, "`"), collapse = ", ")
+    rlang::abort(c(
```
juliasilge, line 117: Should this be `rlang::warn()`? I think probably yes.

juliasilge (changes requested):
Thank you so much for this contribution @brunocarlin!

R/scale.R
brunocarlin: Ok will change it.

R/scale.R
brunocarlin: Ok I can change it back, I will also change 'x' to 'i' since it doesn't stop R anymore

R/scale.R
brunocarlin: I didn't know about that, I should have paid more attention to your original comment my bad. I will implement it.

brunocarlin: Done

R/scale.R
brunocarlin: Done

R/scale.R
brunocarlin: Done

R/scale.R
```diff
@@ -107,9 +107,21 @@ prep.step_scale <- function(x, training, info = NULL, ...) {
 
   sds <-
     vapply(training[, col_names], sd, c(sd = 0), na.rm = x$na_rm)
-
   sds <- sds * x$factor
 
+  which_sd <- which(sds < .Machine$double.eps)
```
topepo, line 112: Can you move this above the `sds <- sds * x$factor` line?

topepo (changes requested):
Thanks for tackling this. 

Can you add a note the NEWS file about this? 

Also, I think that we should edit the `sd` vector to remove the terms with zero(ish) variance and keep the warning. If they are constant then don't do anything to them. 

Leaving it will probably result in a model error (due to NaNs) anyway so I feel that the choices are a controlled halt in the recipe or a warning and no correction.

tests/testthat/test_center_scale_norm.R
```diff
@@ -277,3 +287,10 @@ test_that("normalize - empty printing", {
 
   expect_snapshot(rec)
 })
+
+test_that("normalize - warns on zv",{
+
+  rec1 <- step_normalize(rec_zv,all_numeric_predictors())
+
+  expect_warning(prep(rec1))
+})
```
juliasilge, line 296: Oops, I just noticed this one too. Can you change it to a snapshot test?

brunocarlin:
Should we do what @topepo suggested and just remove those columns? I fear that if someone has some step referencing those by name it may break later.

 We can also add some other value like 0,Inf, or the original value instead of NA but that runs the risk of the user ignoring the warning.

Honestly, after thinking about it for some time the error on prep sounds like the cleanest solution, we can even issue an warning on version 2,* and switch to the error on a later version.

topepo:
Go ahead and remove them since we are not doing anything to them. We do reference that column later to see which columns were affected and removing them would make that more accurate.

brunocarlin:
Just adding

`sds <- sds[-which_sd]`

caused 12 test failings

brunocarlin:
I got a ton of snapshot tests failures don't know yet how to solve those, but the basic behavior seems right to me.

topepo:
No problem. We can edit the PR and take a look at those. I'll work on that this week.

brunocarlin:
I have reduced it to 1 failure, it seems to be with my own test, so it should be fixable.

brunocarlin:
It seems it is actually a bug, Yay to testing will try to solve it.

brunocarlin:
Bug fixed now it will ignore zv on any columns, this causes it to do nothing on both normalize and scale.

R/normalize.R
```diff
@@ -122,6 +122,7 @@ prep.step_normalize <- function(x, training, info = NULL, ...) {
     ))
 
     sds <- sds[-which_sd]
+    means <- means[-which_sd]
```
topepo, line 125: I don't think that this should change. Just because the data cannot be scaled, we can still center them.

brunocarlin:
That will require some checking on the bake part because sweep will break if sds has less elements than means, or we pass sds as 1s

brunocarlin:
I suggest we do it by checking the names in sds

topepo:
@brunocarlin Thanks for getting this going and all of your hard work!

Issue #916, opened 2022-02-24
`step_dummy_multi_choice()` ignores levels of the input factor


zenggyu:
## The problem

`step_dummy_multi_choice()` ignores levels of the input factor and only considers levels in the new data. This causes problem if the downstream model uses a dummy variable for a level that does not occur in the new data. This behavior is in contrast to that of `step_dummy()`.

I also notice that the two functions have a common parameter called `levels`, and it looks like it is intended to explicitly specify the levels to include in the output. However, there are no examples in the documentation and I don't know how to specify (I tried `levels = list(x = c("a", "b", "c", "d", "e", "f", "g"))` but it failed). Can you add some more explanations on how to use the parameter and/or add an example?

## Reproducible example

I would expect the following cases of `step_dummy()` and `step_dummy_multi_choice()` to produce the same number of columns. However, `step_dummy_multi_choice()` does not create columns for levels `"a", "b", "f", "g"`.

By the way, why do `step_dummy()` and `step_dummy_multi_choice()` produce different types of output (double vs integer)?

``` r
suppressPackageStartupMessages(library(recipes))

# old data
tr <- data.frame(x = factor(c("a", "b", "c"), levels = c("a", "b", "c", "d", "e", "f", "g")))

# new data
te <- data.frame(x = factor(c("c", "d", "e"), levels = c("a", "b", "c", "d", "e", "f", "g")))

tr %>%
  recipe() |>
  step_dummy(x, one_hot = T) |>
  prep() |>
  bake(new_data = te)
#> # A tibble: 3 × 7
#>     x_a   x_b   x_c   x_d   x_e   x_f   x_g
#>   <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#> 1     0     0     1     0     0     0     0
#> 2     0     0     0     1     0     0     0
#> 3     0     0     0     0     1     0     0

tr |>
  recipe() |>
  step_dummy_multi_choice(x) |>
  prep() |>
  bake(new_data = te)
#> # A tibble: 3 × 3
#>     x_c   x_d   x_e
#>   <int> <int> <int>
#> 1     1     0     0
#> 2     0     1     0
#> 3     0     0     1
```

<sup>Created on 2022-02-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

EmilHvitfeldt:
[...]

Issue #766, opened 2021-08-11
Have steps return integers when appropriate


EmilHvitfeldt:
Some steps that return what are essentially integers are returning them as doubles. It might be helpful to handle the conversion to help with object sizes.

``` r
library(recipes)
library(modeldata)

data("Chicago")

recipe(ridership ~ date, data = Chicago) %>%
  step_date(date, label = FALSE, keep_original_cols = FALSE) %>%
  prep() %>%
  bake(new_data = NULL)
#> # A tibble: 5,698 × 4
#>    ridership date_dow date_month date_year
#>        <dbl>    <dbl>      <dbl>     <dbl>
#>  1     15.7         2          1      2001
#>  2     15.8         3          1      2001
#>  3     15.9         4          1      2001
#>  4     15.9         5          1      2001
#>  5     15.4         6          1      2001
#>  6      2.42        7          1      2001
#>  7      1.47        1          1      2001
#>  8     15.5         2          1      2001
#>  9     15.9         3          1      2001
#> 10     15.9         4          1      2001
#> # … with 5,688 more rows

recipe(ridership ~ date, data = Chicago) %>%
  step_date(date, keep_original_cols = FALSE) %>%
  step_dummy(all_nominal_predictors()) %>%
  prep() %>%
  bake(new_data = NULL)
#> # A tibble: 5,698 × 19
#>    ridership date_year date_dow_Mon date_dow_Tue date_dow_Wed date_dow_Thu
#>        <dbl>     <dbl>        <dbl>        <dbl>        <dbl>        <dbl>
#>  1     15.7       2001            1            0            0            0
#>  2     15.8       2001            0            1            0            0
#>  3     15.9       2001            0            0            1            0
#>  4     15.9       2001            0            0            0            1
#>  5     15.4       2001            0            0            0            0
#>  6      2.42      2001            0            0            0            0
#>  7      1.47      2001            0            0            0            0
#>  8     15.5       2001            1            0            0            0
#>  9     15.9       2001            0            1            0            0
#> 10     15.9       2001            0            0            1            0
#> # … with 5,688 more rows, and 13 more variables: date_dow_Fri <dbl>,
#> #   date_dow_Sat <dbl>, date_month_Feb <dbl>, date_month_Mar <dbl>,
#> #   date_month_Apr <dbl>, date_month_May <dbl>, date_month_Jun <dbl>,
#> #   date_month_Jul <dbl>, date_month_Aug <dbl>, date_month_Sep <dbl>,
[...]

PR #930, opened 2022-03-11
implement factor levels dummy_multi_choice

brunocarlin:
The aim of this pr is to solve #916 

I haven't been able to do a test without setting up more data, this could be up for debate,

There is also the option of creating a new variable to choose if we keep the behavior of sometimes dropping the "others" column or always wanting the full result. I have coded it to keep with past the behavior where sometimes we drop the "others" column if it is empty


brunocarlin:
There could be also some discussion on dtypes as this returns ints and step_dummy returns doubles

EmilHvitfeldt:
Let us limit the scope of this PR to making sure `step_dummy_multi_choice()` respects factor levels. The dtype issue is already posted here https://github.com/tidymodels/recipes/issues/766. There are some thinking needed to make sure this is a needed change.

EmilHvitfeldt (approved):
It looks good so far. Did a little bit of styling. and removed the new pipe since it would break the CI tests on older R versions.

I'll merge if we have clean GHA checks.

Issue #674, opened 2021-03-27
Keep original labels in `step_discretize`


renanxcortes:
Hi there!

I'd like to ask for a feature that would keep the original labels generated by the internal `cut` function in discretize, instead of "bin1", bin2", etc. Perhaps adding an argument `keep_cut_labels = TRUE`, for example.

Minimal Reproducible Example:

Current Behaviour:

```
library(modeldata)
data(biomass)

biomass_tr <- biomass[biomass$dataset == "Training",]
biomass_te <- biomass[biomass$dataset == "Testing",]

rec <- recipe(HHV ~ carbon,
              data = biomass_tr) %>% 
  step_discretize(carbon)

rec <- prep(rec, biomass_tr)
binned_te <- bake(rec, biomass_te)
table(binned_te$carbon)
```

![image](https://user-images.githubusercontent.com/22593188/112722861-8b919f00-8eea-11eb-8283-f5f167622dec.png)

Expected behaviour:

```
breaks <- quantile(biomass_tr$carbon, probs = seq(0, 1, length = 4 + 1))
table(cut(biomass_te$carbon, breaks = breaks))
```

![image](https://user-images.githubusercontent.com/22593188/112723000-330ed180-8eeb-11eb-9ed1-dd4405bc33e6.png)

juliasilge:
You can get those values out using the `tidy()` method for a recipe:

``` r
library(recipes)
#> Loading required package: dplyr
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
#> 
#> Attaching package: 'recipes'
#> The following object is masked from 'package:stats':
#> 
#>     step
data(biomass, package = "modeldata")

biomass_tr <- biomass[biomass$dataset == "Training",]
biomass_te <- biomass[biomass$dataset == "Testing",]

rec <- recipe(HHV ~ carbon,
              data = biomass_tr) %>% 
  step_discretize(carbon)

rec <- prep(rec, biomass_tr)
binned_te <- bake(rec, biomass_te)
table(binned_te$carbon)
#> 
#> bin_missing        bin1        bin2        bin3        bin4 
#>           0          22          17          25          16

tidy(rec, 1)
#> # A tibble: 5 x 3
#>   terms   value id              
#>   <chr>   <dbl> <chr>           
[...]

Issue #929, opened 2022-03-10
Condition `step_percentile()` on another feature, add `keep_original_cols` arg


joeycouse:
It would be great to have to the ability to condition the use of `step_percentile()` on another feature(s), and have the option to keep the output of the step using `keep_original_cols` which is present in many other step functions.

In the example below, the relationship between time and distance is bimodal, and I would like to filter out the observations that have a much higher distance to time relationship. This creates a much more accurate parameter estimate for the model.

``` r
library(tidymodels)
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip
library(tidyverse)

set.seed(24)

race_data <- tibble(distance = sample(c(2,3,4), size = 1000, replace = T))

race_data <- 
  race_data |> 
  rowwise() |> 
  mutate(final_time = sample(c(rnorm(1, mean = 3, sd = 0.1) * distance,
                             rnorm(1, mean = 7, sd = 0.1) * distance),
                             1,
                             prob = c(0.95,0.05))
         )


# What would be awesome to do inside a reciepe
race_filtered <-
  race_data |> 
  group_by(distance) |> 
  mutate(percentile = ecdf(final_time)(final_time)) |> 
  filter(percentile < 0.90)


# Coefficent Estimate is skewed higher because of outlier points
lm(final_time ~ distance, data = race_data)
#> 
#> Call:
#> lm(formula = final_time ~ distance, data = race_data)
#> 
#> Coefficients:
#> (Intercept)     distance  
#>      -0.394        3.352

# Coefficent Estimate is right on
lm(final_time ~ distance, data = race_filtered)
#> 
#> Call:
#> lm(formula = final_time ~ distance, data = race_filtered)
#> 
#> Coefficients:
#> (Intercept)     distance  
#>    -0.02693      2.99872


# An example of the how the API would work

race_rec <- 
  recipe(final_time ~ distance, data = race_data) |> 
  step_percentile(final_time | distance, keep_original_cols = T) |> 
  step_filter(final_time_percentile < 0.90)
```

<sup>Created on 2022-03-10 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

juliasilge:
I like this idea. What do you think @topepo? Add a `prefix` argument like `"percentile"` and also `keep_original_cols = FALSE`? This would change the output of old recipes, which is a bummer.

brunocarlin:
[...]

PR #951, opened 2022-04-13
Keep original labels in step_discretize

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/recipes/issues/674.

It does 2 things
- makes it so when `prefix = NULL` is set in `discretize()`, that the factor levels match what happens in `cut()`. It essentially disable the custom labeling.
- Sets the default value of `prefix` in `discretize()` to `NULL`.

The old behavior is still possible by setting `prefix` to `"bin"`. It think that changing the default is worthwhile since it aligns it better with what happens in the {embed} steps.

``` r
library(recipes)
library(modeldata)
data(biomass)

recipe(HHV ~ carbon, data = biomass) %>% 
  step_discretize(carbon) %>%
  prep() %>%
  bake(new_data = NULL)
#> # A tibble: 536 × 2
#>    carbon        HHV
#>    <fct>       <dbl>
#>  1 (49.7, Inf]  20.0
#>  2 (47.1,49.7]  19.2
#>  3 (47.1,49.7]  18.3
#>  4 (44.7,47.1]  18.2
#>  5 (44.7,47.1]  18.4
#>  6 (44.7,47.1]  18.5
#>  7 (47.1,49.7]  18.7
#>  8 (44.7,47.1]  18.3
#>  9 (47.1,49.7]  18.6
#> 10 (44.7,47.1]  18.9
#> # … with 526 more rows

recipe(HHV ~ carbon, data = biomass) %>% 
  step_discretize(carbon, options = list(prefix = "bin")) %>%
  prep() %>%
  bake(new_data = NULL)
#> # A tibble: 536 × 2
#>    carbon   HHV
#>    <fct>  <dbl>
#>  1 bin4    20.0
#>  2 bin3    19.2
#>  3 bin3    18.3
#>  4 bin2    18.2
#>  5 bin2    18.4
#>  6 bin2    18.5
#>  7 bin3    18.7
#>  8 bin2    18.3
#>  9 bin3    18.6
#> 10 bin2    18.9
#> # … with 526 more rows
```

<sup>Created on 2022-04-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


topepo (commented):
This should also add a `keep_original_cols` and error if the prefix is NULL

EmilHvitfeldt:
Which function should error if `prefix = NULL`?

topepo:
if `(keep_original_cols && is.null(prefix)`, the `bake` method should error.

EmilHvitfeldt:
I don't think that `keep_original_cols` fits for this step. I wrote up something a little longer here https://github.com/tidymodels/recipes/issues/929#issuecomment-1067082383. In essence; so far we only used `keep_original_cols` in steps that generate new columns. This step doesn't do that.

Issue #919, opened 2022-02-25
Proposal: Create design document


EmilHvitfeldt:
I think it would be useful to have a document (a little bit like [parsnip](https://github.com/tidymodels/parsnip/tree/main/R#readme) has) that contains truths about the ways the steps are expected to behave. I hope that this would be helpful in the long run as we see more and more patterns emerge. I am not expecting this document to be exhaustive. These truths could be paired with code snippets and tests, when applicable, for ease of use.

Top of my mind statements are

- Steps should allow empty selections
- steps that add new columns should allow existing columns to stay using `keep_original_cols`
- What happens when you tidy a unprepped step
- make sure all steps have a tidy method [checking code](https://github.com/tidymodels/recipes/issues/730)
- the columns returned by `tidy()` for a step should not change when the recipe is prepped
- Steps that take factor variables should respect levels
- order of arguments
- how `prefix` should be used
- Order of functions in files
- How deprecation of arguments should be handled [example 1](https://github.com/tidymodels/themis/issues/90) [example 2](https://github.com/tidymodels/embed/pull/118)

I'm sure that there are more

EmilHvitfeldt:
There is talk about this somewhere else. I'm gonna close this issue. Part of it as already been done with the standardization of tests

PR #960, opened 2022-04-20
add initial draft of design document

EmilHvitfeldt:
This PR aims to start the work on the design document and close https://github.com/tidymodels/recipes/issues/919.

So far it is a loosely organized list of truths regarding how the package is/should be written. Over time other rules can be added to this document.

Some of these rules may not be completely satisfied in the package itself, and should be fixed once this PR is merged.


EmilHvitfeldt:
I don't think we are going in this direction design-ducument wise, so I'm will close this PR

PR #971, opened 2022-05-06
Add step_time

### LINKED ISSUE #968: step_time for data-time vectors
To extract, hour, min, decimal time etc.

EmilHvitfeldt:
This PR aims to close #968.

It works almost the exact same way as `step_date()` except it doesn't have a `abbr`, `labels` or `orginal` argument.

``` r
library(recipes)
library(lubridate)

examples <- data.frame(
  times = ymd_hms("2022-05-06 10:01:07") +
  hours(1:5) + minutes(1:5) + seconds(1:5)
)

recipe(~ times, examples) %>%
  step_time(all_predictors()) %>%
  prep() %>%
  bake(new_data = NULL)
#> # A tibble: 5 × 4
#>   times               times_hour24 times_minute times_second
#>   <dttm>                     <int>        <int>        <dbl>
#> 1 2022-05-06 11:02:08           11            2            8
#> 2 2022-05-06 12:03:09           12            3            9
#> 3 2022-05-06 13:04:10           13            4           10
#> 4 2022-05-06 14:05:11           14            5           11
#> 5 2022-05-06 15:06:12           15            6           12
```

<sup>Created on 2022-05-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


NEWS.md
```diff
@@ -23,6 +23,7 @@
 
 * Bug was fixed in `step_holiday()` which used to error when it was applied to variable with missing values. (#743)
 
+* `step_time()` has been added that extracts time features such as hour, minuter, or second. (#968)
```
DavisVaughan, line 26:
```suggestion
* `step_time()` has been added that extracts time features such as hour, minute, or second. (#968)
```

R/date.R
```diff
@@ -39,6 +39,9 @@
 #'  remove the original date variables by default. Set `keep_original_cols`
 #'  to `FALSE` to remove them.
 #'
+#'  See [step_time()] if you want to calculate features that are small than
+#'  days.
```
DavisVaughan, line 43:
```suggestion
#'  See [step_time()] if you want to calculate features that are smaller than
#'  days.
```

Then redocument

R/time.R
```diff
@@ -0,0 +1,197 @@
+#' Time Feature Generator
+#'
+#' `step_time` creates a *specification* of a recipe
```
DavisVaughan, line 3:
```suggestion
#' `step_time()` creates a *specification* of a recipe
```

I think all new documentation should use the tidyverse standard of writing functions with the `()`

R/time.R
```diff
@@ -0,0 +1,197 @@
...
+#'  numeric variables.
+#'
+#' @inheritParams step_pca
+#' @inheritParams step_center
+#' @param ... One or more selector functions to choose variables
+#'  for this step. The selected variables should have class
+#'  `POSIXct` or `POSIXlt`. See [selections()] for more details.
+#' @param features A character string that includes at least one
+#'  of the following values: `ampm` (AM or PM), `hour24`, `hour12`,
+#'  `minute`,`second`, `decimal_day`.
```
DavisVaughan, line 14:
```suggestion
#'  `minute`, `second`, `decimal_day`.
```

DavisVaughan, line 14: Decimal day doesn't feel that useful to me, but I do think it should be documented if you want to retain it. I didn't know what  it did until I looked at the implementation

R/time.R
```diff
@@ -0,0 +1,197 @@
+#' Time Feature Generator
+#'
+#' `step_time` creates a *specification* of a recipe
+#'  step that will convert date-time data into one or more factor or
+#'  numeric variables.
```
DavisVaughan, line 5: I think we should make it always return numeric variables - no factors (im not even sure if it returns a factor at all right now?)

R/time.R
```diff
@@ -0,0 +1,197 @@
...
+#'  `minute`,`second`, `decimal_day`.
+#' @param columns A character string of variables that will be
+#'  used as inputs. This field is a placeholder and will be
+#'  populated once [prep()] is used.
+#' @param keep_original_cols A logical to keep the original variables in the
+#'  output. Defaults to `TRUE`.
+#' @template step-return
+#' @family dummy variable and encoding steps
+#' @export
+#' @details Unlike some other steps, `step_time` does *not*
```
DavisVaughan, line 23:
```suggestion
#' @details Unlike some other steps, `step_time()` does *not*
```

R/time.R
```diff
@@ -0,0 +1,197 @@
...
+#' time_values <- bake(time_rec, new_data = examples)
+#' time_values
+#'
+#' tidy(time_rec, number = 1)
+step_time <-
+  function(recipe,
+           ...,
+           role = "predictor",
+           trained = FALSE,
+           features = c("hour24", "minute", "second"),
```
DavisVaughan, line 59: I'm somewhat partial to using just `"hour"` to mean `"hour24"`, as most people will understand that this is what it means. Then only `"hour12"` is special

R/time.R
```diff
@@ -0,0 +1,197 @@
...
+    feat <-
+      c(
+        "ampm",
+        "hour24",
+        "hour12",
+        "minute",
+        "second",
+        "decimal_day"
+      )
+    if (!is_tune(features) & !is_varying(features)) {
```
DavisVaughan, line 73: We still have `varying()` lying around somewhere? 🤢 

I feel no desire to support that in new steps

R/time.R
```diff
@@ -0,0 +1,197 @@
...
+    id = x$id
+  )
+}
+
+#' @export
+bake.step_time <- function(object, new_data, ...) {
+
+  for (column in object$columns) {
+    time_values <- get_time_features(
+      dt = getElement(new_data, column),
```
DavisVaughan, line 143:
```suggestion
      dt = new_data[[column]],
```

Still prefer this 😛

R/time.R
```diff
@@ -0,0 +1,197 @@
...
+    new_data <- new_data[, !(colnames(new_data) %in% object$columns), drop = FALSE]
+  }
+
+  new_data
+}
+
+get_time_features <- function(dt, feats) {
+
+  features <- list(
+    ampm = function(x) format(x, "%p"),
```
DavisVaughan, line 162: Rather than returning a character vector here, I feel like it would be more appropriate to return a logical column that is `TRUE` if AM and `FALSE` if PM. Then you could just use `lubridate::am()` directly.

The logical column feels more useful for modeling? I think? Or maybe a binary integer column?

If you go this route, maybe it should be called `am` rather than `ampm`, or you could just document `ampm` well to ensure that people know which one corresponds to `TRUE`

R/time.R
```diff
@@ -0,0 +1,197 @@
...
+    hour24 = hour,
+    hour12 = function(x) as.integer(format(x, "%I")),
```
DavisVaughan, line 164: Probably faster:

```r
function(x) {
  out <- hour(x)
  out <- ifelse(out > 12L, out - 12L, out)
  out
}
```

tests/testthat/test_time.R
```diff
@@ -0,0 +1,135 @@
...
+
+examples <- data.frame(
+  times = ymd_hms("2022-05-06 10:01:07") +
+  hours(1:5) + minutes(1:5) + seconds(1:5)
+)
+
+date_rec <- recipe(~ times, examples) %>%
+  step_time(all_predictors())
+
+feats <- c("ampm", "hour24", "hour12", "minute", "second", "decimal_day")
```
DavisVaughan, line 15: I think we should enforce Jenny's testing suggestions in all new test files. So

- We can `library(testthat)` for Max, I guess
- No need to `library(recipes)` because even Max will do that
- Definitely avoid `library(lubridate)` and `library(tibble)` as that can have far reaching side effects

I see nothing wrong with duplicating these data objects in each of the tests. If you feel strongly against that, I'd create `helper-time.R` and add something like `time_example()` that makes the data frame, `time_recipe()` that makes the recipe and `time_features()` that make the features vector.

tests/testthat/test_time.R
```diff
@@ -0,0 +1,135 @@
+library(testthat)
+library(recipes)
+library(lubridate)
+library(tibble)
+
```
DavisVaughan, line 5: Noting that the file name should be `test-time.R`, i.e. created with `use_test()`.

tests/testthat/test_time.R
```diff
@@ -0,0 +1,135 @@
...
+  expect_identical(tidy(rec, number = 1), expect)
+
+  rec <- prep(rec, mtcars)
+
+  expect_identical(tidy(rec, number = 1), expect)
+})
+
+test_that("empty printing", {
+  skip_if(packageVersion("rlang") < "1.0.0")
```
DavisVaughan, line 126: I think it is time to require CRAN rlang in Imports. i.e. `rlang (>= 1.0.2)`. There is no reason not to and it makes our testing simpler because we don't need skips

DavisVaughan (changes requested):

R/time.R
EmilHvitfeldt: the `ampm` variables returns a factor with levels "am" and "pm"

EmilHvitfeldt: Will redo `ampm` variable to be 0-1 coded,

R/time.R
EmilHvitfeldt: Good catch, I'll remove it here and add another issue to deal with the remaining. Use of varying was deprecated back in 0.1.17

R/time.R
EmilHvitfeldt: keeping it, added documentation

R/time.R
EmilHvitfeldt: changing to am()

tests/testthat/test_time.R
EmilHvitfeldt: Thank you, done!

PR #975, opened 2022-05-10
reduce use of sweep() in steps

EmilHvitfeldt:
This PR is very similar to https://github.com/tidymodels/recipes/pull/972.

The `sweep()`s in `pls_project()` and `old_pls_project()` are kept in because the data are already matrices and we don't gain any improvement from removing them since we need to do `%*%`.

### DIFF
--- a/R/scale.R
+++ b/R/scale.R
@@ -124,11 +124,11 @@ prep.step_scale <- function(x, training, info = NULL, ...) {
 
 #' @export
 bake.step_scale <- function(object, new_data, ...) {
-  res <-
-    sweep(as.matrix(new_data[, names(object$sds)]), 2, object$sds, "/")
-  res <- tibble::as_tibble(res)
-  new_data[, names(object$sds)] <- res
-  as_tibble(new_data)
+  for (column in names(object$sds)) {
+    sd <- object$sds[column]
+    new_data[[column]] <- new_data[[column]] / sd
+  }
+  new_data
 }
 
 print.step_scale <-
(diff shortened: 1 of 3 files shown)


R/scale.R
```diff
@@ -124,11 +124,11 @@ prep.step_scale <- function(x, training, info = NULL, ...) {
...
-  res <-
-    sweep(as.matrix(new_data[, names(object$sds)]), 2, object$sds, "/")
-  res <- tibble::as_tibble(res)
-  new_data[, names(object$sds)] <- res
-  as_tibble(new_data)
+  for (column in names(object$sds)) {
+    sd <- object$sds[column]
+    new_data[[column]] <- new_data[[column]] / sd
+  }
+  new_data
```
simonpcouch, line 131: I think we may still want to return `tibble::as_tibble(new_data)` in these spots unless there's some machinery that processes `new_data` beforehand and we can guarantee `new_data` is a tibble.

i.e. reading through this, I'm not sure we can guarantee we'll return a tibble—acknowledging that I see the removal of the `as.matrix` line.🙂

simonpcouch (commented):
Just have one thing I want to confirm here, otherwise looks good to me. :)

Hadn't seen `sweep` before—what a gem!😬

simonpcouch (approved):
Oh, my comment looks related to #979. LGTM. :)

R/scale.R
EmilHvitfeldt: Sweep isn't actually that bad, it just happens to slow us down here since we need to `as.matrix()` and `as_tibble()`.

We don't need to ensure that we get a tibble since that is done automatically after each bake step is being performed https://github.com/tidymodels/recipes/blob/77fa788b3fed542c7832fc00a8b9725a165e6a46/R/recipe.R#L580.

https://github.com/tidymodels/recipes/pull/979 I have a separate PR to remove the duplicate instances of `as_tibble()`.

PR #976, opened 2022-05-10
remove unused functions

### LINKED ISSUE #853: unused functions in misc.R
These functions appear to no longer be used.

https://github.com/tidymodels/recipes/blob/558625689b0b4c87761cd934acc3972ff5e581e6/R/misc.R#L75-L76

EmilHvitfeldt:
These 3 functions are not used nor are they exported and are thus being deleted.

To close #853

### DIFF
--- a/R/tunable.R
+++ b/R/tunable.R
@@ -16,8 +16,6 @@ tunable.step <- function(x, ...) {
   no_param
 }
 
-step_type <- function(.step) class(.step)[class(.step) != "step"][1]
-
 no_param <-
   tibble::tibble(
     name = NA_character_,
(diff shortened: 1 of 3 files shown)


R/tunable.R
```diff
@@ -16,8 +16,6 @@ tunable.step <- function(x, ...) {
   no_param
 }
 
-step_type <- function(.step) class(.step)[class(.step) != "step"][1]
```
simonpcouch, line 19: Is this relevant here?

https://github.com/tidymodels/recipes/blob/8f5afe1/tests/testthat/test_skipping.R#L44

simonpcouch (commented):
Just left one comment. Otherwise, LGTM. :)

R/tunable.R
EmilHvitfeldt: Yes, we can remove that line as well, good catch!

Issue #811, opened 2021-09-18
Moving presentation folder


EmilHvitfeldt:
The [presentations](https://github.com/tidymodels/recipes/tree/master/presentations) folder contains a single presentation from 2017. Should this be moved somewhere else?

PR #978, opened 2022-05-10
delete old presentation

EmilHvitfeldt:
This PR will close https://github.com/tidymodels/recipes/issues/811 by deleting the presentation.


hfrick:
I agree that this should be removed from recipes! but 2017 clearly predates me as well, so: @topepo is that your presentation? should it be moved somewhere else or just deleted?

hfrick:
I agree that this should be removed from recipes! but 2017 clearly predates me as well, so: @topepo is that your presentation? should it be moved somewhere else or just deleted?

EmilHvitfeldt:
The presentation will still be able to be recovered if we merge this PR, you just have to go back to an earlier commit, which we can find by search "presentation" which brings up this PR 😄

hfrick:
right, I was just thinking/remembering the same 👍

hfrick (approved):

Issue #798, opened 2021-09-14
use of as_tibble in bake.*() functions


EmilHvitfeldt:
It is inconsistent how `bake.*()` steps are ended, with respect to keeping `new_data` as a tibble. 

Some steps like `step_range()` just returns `as_tibble(new_data)`

https://github.com/tidymodels/recipes/blob/788d409b96587a3377b5846df93774dfce511726/R/range.R#L123

other steps like `step_integer()`, tests before,

https://github.com/tidymodels/recipes/blob/788d409b96587a3377b5846df93774dfce511726/R/integer.R#L153-L155

Lastly, there are steps that doesn't end with a call to `as_tibble()` such as `step_relu()`

https://github.com/tidymodels/recipes/blob/788d409b96587a3377b5846df93774dfce511726/R/relu.R#L143-L151

DavisVaughan:
I don't think it ends up mattering too much since after each bake step we convert to a tibble
https://github.com/tidymodels/recipes/blob/77fa788b3fed542c7832fc00a8b9725a165e6a46/R/recipe.R#L580

EmilHvitfeldt:
Neat, then it seems that `as_tibble()` calls inside bake steps are rather superfluous.

PR #979, opened 2022-05-10
don't end with as_tibble() in bake functions

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/recipes/issues/798

Since `bake.recipe()` calls `as_tibble()` after each step we don't need to do it inside each step.


hfrick (approved):
Nice! 🧹 ✨

PR #984, opened 2022-05-18
Add missing skip if not installed

EmilHvitfeldt:
This PR adds `skip_if_not_installed()` to a couple of test files that didn't have them for steps that uses suggested packages.


simonpcouch (approved):
Yup, looks good to me! 

I've found the "hard" check Action, that doesn't use `noSuggests`, to be a nice way to more rigorously check conditional usage.

https://github.com/tidymodels/broom/blob/main/.github/workflows/R-CMD-check-hard.yaml

PR #992, opened 2022-05-25
Checks to make sure bake steps only returns tibbles

EmilHvitfeldt:
This PR does 2 things, from internal discussion, it changes the following chunks in `bake.recipe()` and `prep.recipe()`

```r
if (!is_tibble(training)) {
  training <- as_tibble(training)
}
```

to throw an error if the output of a `bake.step_*()` doesn't return a tibble. Luckily none of the steps in this package had this behavior 😅 

In addition it adds the un-exported `step_testthat_helper()` step that is used to test this feature. Right now it is used as a testing feature. It is only used in one set of tests right now, but it can easily be adapted and used in other tests for bad design.


topepo:
Tag this PR when you make the tidymodls.org change. I'll merge when that is ready.

PR #996, opened 2022-05-26
add R-CMD-check-hard GHA

EmilHvitfeldt:
As discussed in our meeting today, this mimics the CRAN NoSuggests environment.


juliasilge (approved):
This looks great! 🙌 

The only caution I have is that I got vetiver to pass with `R-CMD-check-hard`, but it [turns out there were _more_ packages I should have skipped](https://github.com/rstudio/vetiver-r/pull/96). The situation was that it was skipping the tests because some packages were not installed, but this check doesn't robustly check that test will skip with any combination of packages missing.

I spent some time looking for any occurrences of this in the tests but didn't find any. It might be worth one more look by you to check for this.

EmilHvitfeldt:
That is a good note. Did another round and couldn't find anything

Issue #989, opened 2022-05-24
action with strings2factor


topepo:
Based on [this community issue](https://community.rstudio.com/t/error-with-step-string2factor-and-model-global-explanations/137939/7) (and previous interactions), we should caution users not to use this step and instead use `across` before any interactions with tidymodels.

PR #997, opened 2022-05-26
Strings2factor

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/recipes/issues/989 by adding some words of caution to the `step_string2factor()` documentation.


R/string2factor.R
```diff
@@ -12,7 +12,22 @@
 #' @template step-return
 #' @family dummy variable and encoding steps
 #' @export
-#' @details If `levels` is given, `step_string2factor` will
+#' @details
+#'  In most cases, if you are planning to use `step_string2factor()`
+#'  without setting `levels`, you will be better off by converting
```
juliasilge, line 17:
```suggestion
#'  without setting `levels`, you will be better off converting
```

R/string2factor.R
```diff
@@ -12,7 +12,22 @@
 #' @template step-return
 #' @family dummy variable and encoding steps
 #' @export
-#' @details If `levels` is given, `step_string2factor` will
+#' @details
+#'  In most cases, if you are planning to use `step_string2factor()`
```
topepo, line 16: Can you add a top-level statement that tells people to only use this in special cases and, instead, set the levels before using any tidymodels functions? That's the message and it is buried in a lot of minutiae.

juliasilge: Do you mean in the description? That seems like a good addition to me to what is already in the details.

R/string2factor.R
```diff
@@ -1,7 +1,9 @@
 #' Convert Strings to Factors
 #'
 #' `step_string2factor` will convert one or more character
-#'  vectors to factors (ordered or unordered).
+#'  vectors to factors (ordered or unordered). This step should only be
+#'  used in special cases. Factors should generally be created before using
+#'  any tidymodels functions.
```
juliasilge, line 6:
```suggestion
#'  vectors to factors (ordered or unordered). Use this step only 
#'  in special cases (see Details) and instead convert strings to 
#'  factors before using any tidymodels functions.
```

PR #865, opened 2021-12-10
error in bake() if variable is missing in new_data

EmilHvitfeldt:
This PR aims to close #491.

It adds a guard in `bake.recipe()` that checks that all the original predictors are available. Errors out otherwise, consistent with similar errors such as from `check_cols()`.

A couple of existing tests were changed to adhere to this higher level of strictness 

# Example:

``` r
library(recipes)
library(modeldata)

data(ames)
rec <- recipe(Sale_Price ~ ., data = ames) %>%
  step_dummy(MS_SubClass, MS_Zoning)
prepped <- prep(rec)

bake(prepped, ames %>% dplyr::select(-MS_SubClass))
#> Error: The following cols are missing from `new_data`: MS_SubClass
bake(prepped, ames %>% dplyr::select(-MS_SubClass, -MS_Zoning))
#> Error: The following cols are missing from `new_data`: MS_SubClass, MS_Zoning
bake(prepped, ames %>% dplyr::select(Sale_Price))
#> Error: The following cols are missing from `new_data`: MS_SubClass, MS_Zoning...
```

PR #999, opened 2022-06-06
more documentation for non-standard roles

topepo:
Related to #865 as well as tidymodels/tune#494

Is there anything else to add here (perhaps related to `bake()`)? I've mostly mentioned things related to tune or workflows.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,7 @@ Imports:
     generics (>= 0.1.0.9000),
     glue,
     gower,
-    hardhat (>= 0.1.6.9001),
+    hardhat (>= 1.0.0),
     ipred (>= 0.9-12),
     lifecycle,
     lubridate (>= 1.8.0),
@@ -61,12 +61,10 @@ Suggests:
     xml2
 VignetteBuilder: 
     knitr
-Remotes: 
-    tidymodels/hardhat
 RdMacros: 
     lifecycle
 Config/Needs/website: tidyverse/tidytemplate
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.0
+RoxygenNote: 7.2.0.9000
--- a/man/rmd/bake_dependent_roles.Rmd
+++ b/man/rmd/bake_dependent_roles.Rmd
@@ -1,7 +1,7 @@
 ## Effects of non-standard roles
 
 
-Recipes can label and retain column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development.
+Recipes can label and retain column(s) of your data set that should not be treated as outcomes or predictors. A unique identifier column or some other ancillary data could be used to troubleshoot issues during model development but may not be either an outcome or predictor.
 
 As an example, you might have a date column that is transformed into features using [step_date()] or [step_holiday()]. 
 
--- a/man/rmd/bake_dependent_roles.Rmd
+++ b/man/rmd/bake_dependent_roles.Rmd
@@ -1,9 +1,9 @@
 ## Effects of non-standard roles
 
 
-Recipes steps can keep column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development. 
+Recipes can label and retain column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development.
 
-As an example, you might have a date column that is transformed into features using [step_date()] and.or [step_holiday()]. 
+As an example, you might have a date column that is transformed into features using [step_date()] or [step_holiday()]. 
 
 ```{r, results = "hide", message = FALSE, warning = FALSE}
   library(tidymodels)
@@ -17,9 +17,9 @@ As an example, you might have a date column that is transformed into features us
     step_date(date)
 ```
 
-At this pont, `date` is kept in the dataset as-is, and will be treated as a predictor (and will probably be converted to an integer column by the modeling function). 
+At this point, `date` is kept in the dataset as-is, and will be treated as a predictor (and will probably be converted to an integer column by the modeling function). 
 
-One option is to remove date since it is not directly used as a model term. [step_rm()] is a good way to accomplish this. 
+One option is to remove `date` since it is not directly used as a model term. [step_rm()] is a good way to accomplish this. 
 
 ```{r}
   chi_rec_rm <- chi_rec %>% step_rm(date)
@@ -31,11 +31,11 @@ Another approach, especially helpful during model development, is to change the 
   chi_rec_role <- chi_rec %>% update_role(date, new_role = "date")
 ```
 
-In this case, the column is still available in the data and will be carried through the analysis (e.g. during resampling). This can be helpful since, it there is a poor prediction, this column may be helpful in diagnosing the problem. 
+In this case, the column is still available in the data and will be carried through the analysis (e.g. during resampling). This can be helpful since, if there is a poor prediction, this column may be helpful in diagnosing the problem. 
 
 The downside to chaining the role of the column is that it is needed at prediction-time. This can impact development tools (such as tuning functions) and would also be required to be in the data being predicted once the model is deployed. 
 
-If using a non-standard role for a column, it highly recommended that you use the recipe within a workflow and specify `bake_dependent_roles` as described in [hardhat::default_recipe_blueprint()]. This will allow tuning and prediction functions to run without error. 
+If using a non-standard role for a column, it is highly recommended that you use the recipe within a workflow and specify `bake_dependent_roles` as described in [hardhat::default_recipe_blueprint()]. This will allow tuning and prediction functions to run without error. 
 
 With the example above:
 
--- a/man/rmd/bake_dependent_roles.Rmd
+++ b/man/rmd/bake_dependent_roles.Rmd
@@ -3,7 +3,18 @@
 
 Recipes can label and retain column(s) of your data set that should not be treated as outcomes or predictors. A unique identifier column or some other ancillary data could be used to troubleshoot issues during model development but may not be either an outcome or predictor.
 
-As an example, you might have a date column that is transformed into features using [step_date()] or [step_holiday()]. 
+For example, the [modeldata::biomass] data have a column called `sample` with information about the specific sample type. We can change that role: 
+
+```{r, eval = FALSE}
+  data(biomass, package = "modeldata")
+
+  recipe(HHV ~ ., data = biomass) %>%
+    update_role(sample, new_role = "id variable")
+```
+
+This means that the `sample` column will not be used in the analysis but will be retained in the data set (see more in the Examples section below). 
+
+As another example, you might have a date column that is transformed into features using [step_date()] or [step_holiday()]. 
 
 ```{r, results = "hide", message = FALSE, warning = FALSE}
   library(tidymodels)
@@ -53,3 +64,10 @@ With the example above:
   # This line will fail without the added blueprint:
   predict(chi_fit, head(Chicago))
 ```
+
+It is also worth noting that [workflows::augment.workflow()] can be used to attach predictions to the data (including columns with non-standard roles):
+
+```{r, eval = FALSE}
+  augment(chi_fit, head(Chicago))
+```
+
--- a/man/rmd/bake_dependent_roles.Rmd
+++ b/man/rmd/bake_dependent_roles.Rmd
@@ -1,0 +1,55 @@
+## Effects of non-standard roles
+
+
+Recipes steps can keep column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development. 
+
+As an example, you might have a date column that is transformed into features using [step_date()] and.or [step_holiday()]. 
+
+```{r, results = "hide", message = FALSE, warning = FALSE}
+  library(tidymodels)
+  tidymodels_prefer()
+
+  data(Chicago, package = "modeldata")
+  chi_rec <- recipe(ridership ~ ., data = Chicago) %>% 
+    # Since `date` is symbolically on the right-hand side of the formula, it
+    # is treated as a predictor (i.e., has a role of "predictor")
+    step_holiday(date) %>% 
+    step_date(date)
+```
+
+At this pont, `date` is kept in the dataset as-is, and will be treated as a predictor (and will probably be converted to an integer column by the modeling function). 
+
+One option is to remove date since it is not directly used as a model term. [step_rm()] is a good way to accomplish this. 
+
+```{r}
+  chi_rec_rm <- chi_rec %>% step_rm(date)
+```
+
+Another approach, especially helpful during model development, is to change the role of the column to some other value via [update_role()]. 
+
+```{r}
+  chi_rec_role <- chi_rec %>% update_role(date, new_role = "date")
+```
+
+In this case, the column is still available in the data and will be carried through the analysis (e.g. during resampling). This can be helpful since, it there is a poor prediction, this column may be helpful in diagnosing the problem. 
+
+The downside to chaining the role of the column is that it is needed at prediction-time. This can impact development tools (such as tuning functions) and would also be required to be in the data being predicted once the model is deployed. 
+
+If using a non-standard role for a column, it highly recommended that you use the recipe within a workflow and specify `bake_dependent_roles` as described in [hardhat::default_recipe_blueprint()]. This will allow tuning and prediction functions to run without error. 
+
+With the example above:
+
+```{r, warning = FALSE}
+  library(hardhat) # require hardhat version >= 1.0.0
+  # Make a new blueprint:
+  date_bp <- default_recipe_blueprint(bake_dependent_roles = "date")
+
+  chi_wflow <- 
+    workflow() %>% 
+    add_model(linear_reg()) %>% 
+    add_recipe(chi_rec_role, blueprint = date_bp) 
+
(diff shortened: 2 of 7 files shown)


man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
...
+  chi_rec <- recipe(ridership ~ ., data = Chicago) %>% 
+    # Since `date` is symbolically on the right-hand side of the formula, it
+    # is treated as a predictor (i.e., has a role of "predictor")
+    step_holiday(date) %>% 
+    step_date(date)
+```
+
+At this pont, `date` is kept in the dataset as-is, and will be treated as a predictor (and will probably be converted to an integer column by the modeling function). 
+
+One option is to remove date since it is not directly used as a model term. [step_rm()] is a good way to accomplish this.
```
EmilHvitfeldt, line 22:
```suggestion
One option is to remove `date` since it is not directly used as a model term. [step_rm()] is a good way to accomplish this. 
```

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
...
+  chi_rec_rm <- chi_rec %>% step_rm(date)
+```
+
+Another approach, especially helpful during model development, is to change the role of the column to some other value via [update_role()]. 
+
+```{r}
+  chi_rec_role <- chi_rec %>% update_role(date, new_role = "date")
+```
+
+In this case, the column is still available in the data and will be carried through the analysis (e.g. during resampling). This can be helpful since, it there is a poor prediction, this column may be helpful in diagnosing the problem.
```
EmilHvitfeldt, line 34:
```suggestion
In this case, the column is still available in the data and will be carried through the analysis (e.g. during resampling). This can be helpful since, if there is a poor prediction, this column may be helpful in diagnosing the problem. 
```

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
...
+
+```{r}
+  chi_rec_role <- chi_rec %>% update_role(date, new_role = "date")
+```
+
+In this case, the column is still available in the data and will be carried through the analysis (e.g. during resampling). This can be helpful since, it there is a poor prediction, this column may be helpful in diagnosing the problem. 
+
+The downside to chaining the role of the column is that it is needed at prediction-time. This can impact development tools (such as tuning functions) and would also be required to be in the data being predicted once the model is deployed. 
+
+If using a non-standard role for a column, it highly recommended that you use the recipe within a workflow and specify `bake_dependent_roles` as described in [hardhat::default_recipe_blueprint()]. This will allow tuning and prediction functions to run without error.
```
EmilHvitfeldt, line 38:
```suggestion
If using a non-standard role for a column, it is highly recommended that you use the recipe within a workflow and specify `bake_dependent_roles` as described in [hardhat::default_recipe_blueprint()]. This will allow tuning and prediction functions to run without error. 
```

DESCRIPTION
```diff
@@ -61,12 +61,10 @@ Suggests:
...
     knitr
-Remotes: 
-    tidymodels/hardhat
 RdMacros: 
     lifecycle
 Config/Needs/website: tidyverse/tidytemplate
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.0
```
DavisVaughan, line 72: Is there a reason you have dev roxygen2? 7.2.0 just came out!

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
+## Effects of non-standard roles
+
+
+Recipes steps can keep column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development. 
+
+As an example, you might have a date column that is transformed into features using [step_date()] and.or [step_holiday()]. 
+
+```{r, results = "hide", message = FALSE, warning = FALSE}
+  library(tidymodels)
```
DavisVaughan, line 9: It looks like all of the code in your code chunks are indented, was that on purpose?

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
+## Effects of non-standard roles
+
+
+Recipes steps can keep column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development.
```
DavisVaughan, line 4:
```suggestion
Recipes can label and retain column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development.
```

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
+## Effects of non-standard roles
+
+
+Recipes steps can keep column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development. 
+
+As an example, you might have a date column that is transformed into features using [step_date()] and.or [step_holiday()].
```
DavisVaughan, line 6:
```suggestion
As an example, you might have a date column that is transformed into features using [step_date()] or [step_holiday()]. 
```

DavisVaughan (changes requested):
I think I have some larger comments that I'd like addressed in some way first before digging into the typos and smaller things.

I'd really like to push people towards `step_rm(x)` if they use `x` to derive any other features - which is the case here with `date`. I'd say that `date` _is_ a predictor, because it is used to create other features, so assigning it a non-standard role feels very wrong.

I no longer believe that roles outside of `"outcome"`, `"predictor"`, and `"case_weights"` have that much value. If you are going to assign something a non-standard role, I think that the only time that is acceptable is if that column isn't used in ANY of the steps. But that means that that is essentially equivalent to removing it from the data up front, so why not just do that?

If you are going to keep usage of `update_role()` in here, then I think it is important to mention that `update_role()` happens before any of the steps have run, since that is so hard to remember.

I'm not sure that I agree with your example about keeping `date` around as a `"date"` role just to help diagnose a problem. We teach and recommend that people use a workflow for fitting their models, which means that most people will never call `bake()` themselves, which means they will never see `date` carried through. I feel like if you wanted to mention this then it could be a section at the very end that is essentially like "There are very few examples where using a custom role is useful. That said, one example of this is...".

Right now it just feels like we are saying that `step_rm()` and `update_role()` are equal in terms of usefulness, and I no longer feel like that is the case.

EmilHvitfeldt:
> I'd really like to push people towards step_rm(x) if they use x to derive any other features - which is the case here with date. I'd say that date is a predictor, because it is used to create other features, so assigning it a non-standard role feels very wrong.

I agree that we should think of this data column as a predictor, and short term `step_rm()` works really well here. As a larger point I think this issue pops up because of the inconsistent use of `keep_original_cols` in the following steps which IMO should have `keep_original_cols = FALSE` like the other steps that has `keep_original_cols`.

- `step_time()`
- `step_date()`
- `step_holiday()`
- `step_ratio()`

I find this recipe ideal

```r
 recipe(ridership ~ ., data = Chicago) %>% 
    step_holiday(date, keep_original_cols = TRUE) %>% 
    step_date(date)
```

compared to these two, preferring the use of `keep_original_cols` over `step_rm()`.

```r
 recipe(ridership ~ ., data = Chicago) %>% 
    step_holiday(date) %>% 
    step_date(date, keep_original_cols = FALSE)

 recipe(ridership ~ ., data = Chicago) %>% 
    step_holiday(date) %>% 
    step_date(date) %>%
    step_rm(date)
```

> I'm not sure that I agree with your example about keeping date around as a "date" role just to help diagnose a problem.

I think having `augment()` to use on a workflow alleviates the need keep variables around for diagnosis anyways.

juliasilge:
I think that using date here is not a great example for teaching/explanation, because with this data I would recommend (like Emil) to set up the steps to use `keep_original_cols = FALSE`. We are moving toward all our steps that create new variables having this option. 

There are a few examples where I have used other roles that are IMO better use cases:

- https://juliasilge.com/blog/cocktail-recipes-umap/
- https://juliasilge.com/blog/board-games/ (don't actually use it in this, but I did in the video, I think?)

Showing how to use the other roles for a totally separate variable is a better call, for visualization or diagnostics, etc. A possible example would be [`tate_text`, and show how to use title and ID](https://modeldata.tidymodels.org/dev/reference/tate_text.html) as "other" roles.

EmilHvitfeldt:
> Showing how to use the other roles for a totally separate variable is a better call, for visualization or diagnostics, etc. A possible example would be [tate_text, and show how to use title and ID](https://modeldata.tidymodels.org/dev/reference/tate_text.html) as "other" roles.

Isn't this what we are trying to tell people not to do? What would be the benefit of doing this compared to using `augment()`?

juliasilge:
OK, I had a misunderstanding. I did not know this worked (i.e. you can augment on a dataframe with _anything_ as long as it has the predictors):

``` r
library(tidymodels)
data(tate_text)

tree_spec <- decision_tree(mode = "regression")
rec <- recipe(year ~ artist + medium, data = tate_text %>% select(year, artist, medium))

new_data <- tibble(id = 333, artist = "Baselitz, Georg", title = "A beautiful painting!", medium = "Oil paint on canvas")

workflow(rec, tree_spec) %>% fit(data = tate_text) %>% predict(new_data)
#> # A tibble: 1 × 1
#>   .pred
#>   <dbl>
#> 1 1995.
workflow(rec, tree_spec) %>% fit(data = tate_text) %>% augment(new_data)
#> # A tibble: 1 × 5
#>      id artist          title                 medium              .pred
#>   <dbl> <chr>           <chr>                 <chr>               <dbl>
#> 1   333 Baselitz, Georg A beautiful painting! Oil paint on canvas 1995.
```

<sup>Created on 2022-06-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This does change my opinion a bit. Now I think:

- the date example is not really great because it would be better to handle that with recipe steps + recipe step args
- `augment()` is better for diagnosing workflows
- the only "good" use for the "other" roles I can think of now is post-baking diagnosing of standalone recipes (not in a workflow)

EmilHvitfeldt:
I fully agree with the following:

> This does change my opinion a bit. Now I think:
> 
> * the date example is not really great because it would be better to handle that with recipe steps + recipe step args
> * `augment()` is better for diagnosing workflows
> * the only "good" use for the "other" roles I can think of now is post-baking diagnosing of standalone recipes (not in a workflow)

Should be allow for augment to work on recipes??

juliasilge:
I like that plan a lot. I think that means instead of having a separate vignette, we only update the `update_role()` docs. I would say also update the `augment()` docs to show more clearly having "other" columns and how to use for diagnostics.

DESCRIPTION
topepo: We tend to use the dev version as much as possible.

DavisVaughan: That sounds like a painful way to live 😅

man/rmd/bake_dependent_roles.Rmd
topepo: it is!

DESCRIPTION
```diff
@@ -24,7 +24,7 @@ Imports:
     generics (>= 0.1.0.9000),
     glue,
     gower,
-    hardhat (>= 0.1.6.9001),
+    hardhat (>= 1.0.0),
```
juliasilge, line 27: This should be 1.1.0 now, right?

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
+## Effects of non-standard roles
+
+
+Recipes can label and retain column(s) of your data set that should not be treated as outcomes or predictors. For example, a unique identifier column or some other ancillary data that can be used to troubleshoot issues during model development.
```
juliasilge, line 4:
```suggestion
Recipes can label and retain column(s) of your data set that should not be treated as outcomes or predictors. A unique identifier column or some other ancillary data could be used to troubleshoot issues during model development but may not be either an outcome or predictor.
```

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
...
+```{r, results = "hide", message = FALSE, warning = FALSE}
+  library(tidymodels)
+  tidymodels_prefer()
+
+  data(Chicago, package = "modeldata")
+  chi_rec <- recipe(ridership ~ ., data = Chicago) %>% 
+    # Since `date` is symbolically on the right-hand side of the formula, it
+    # is treated as a predictor (i.e., has a role of "predictor")
+    step_holiday(date) %>% 
+    step_date(date)
```
juliasilge, line 17: Can we use something here that is not `date`? I think this gets really muddy because it is sort of a predictor and you are using it to create predictors, and we have better ways to not keep the original variable around. I think `tate_text` would be a better option for these docs.

man/rmd/recipes.Rmd
```diff
@@ -22,13 +22,13 @@ There are two different types of operations that can be sequentially added to a
 
 - **Checks** are operations that conduct specific tests of the data. When the test is satisfied, the data are returned without issue or modification. Otherwise, an error is thrown.
 
-If you have defined a recipe and want to see which steps are included, use the `tidy()` method on the recipe object.
+If you have defined a recipe and want to see which steps are included, use the [tidy()] method on the recipe object.
```
juliasilge, line 25:
```suggestion
If you have defined a recipe and want to see which steps are included, use the [`tidy()`][tidy.recipe()] method on the recipe object.
```
Otherwise it links to something confusing

man/rmd/recipes.Rmd
```diff
@@ -73,13 +73,13 @@ linear_sp_sign_wflow <-
 linear_sp_sign_wflow
 ```
 
-To estimate the preprocessing steps and then fit the linear model, a single call to `fit()` is used:
+To estimate the preprocessing steps and then fit the linear model, a single call to [fit()] is used:
```
juliasilge, line 76:
```suggestion
To estimate the preprocessing steps and then fit the linear model, a single call to [`fit()`][fit.model_spec()] is used:
```

man/rmd/recipes.Rmd
```diff
@@ -73,13 +73,13 @@ linear_sp_sign_wflow <-
...
 
 ```{r}
 linear_sp_sign_fit <- fit(linear_sp_sign_wflow, data = biomass_tr)
 ```
 
-When predicting, there is no need to do anything other than call `predict()`. This preprocesses the new data in the same manner as the training set, then gives the data to the linear model prediction code: 
+When predicting, there is no need to do anything other than call [predict()]. This preprocesses the new data in the same manner as the training set, then gives the data to the linear model prediction code:
```
juliasilge, line 82:
```suggestion
When predicting, there is no need to do anything other than call [`predict()`][predict.model_fit()]. This preprocesses the new data in the same manner as the training set, then gives the data to the linear model prediction code: 
```

man/rmd/recipes.Rmd
```diff
@@ -116,9 +116,9 @@ You can [tidy.recipe()] a recipe, either when it is prepped or unprepped, to lea
 tidy(pca_rec)
 ```
 
-You can also `tidy()` recipe *steps* with a `number` or `id` argument.
+You can also [tidy()] recipe *steps* with a `number` or `id` argument.
```
juliasilge, line 119:
```suggestion
You can also [`tidy()`][tidy.recipe()] recipe *steps* with a `number` or `id` argument.
```

man/rmd/recipes.Rmd
```diff
@@ -116,9 +116,9 @@ You can [tidy.recipe()] a recipe, either when it is prepped or unprepped, to lea
...
 
-To apply the prepped recipe to a data set, the [bake()] function is used in the same manner that `predict()` would be for models. This applies the estimated steps to any data set.
+To apply the prepped recipe to a data set, the [bake()] function is used in the same manner that [predict()] would be for models. This applies the estimated steps to any data set.
```
juliasilge, line 121:
```suggestion
To apply the prepped recipe to a data set, the [bake()] function is used in the same manner that [`predict()`][predict.model_fit()] would be for models. This applies the estimated steps to any data set.
```

man/rmd/bake_dependent_roles.Rmd
```diff
@@ -0,0 +1,55 @@
...
+
+```{r}
+  chi_rec_role <- chi_rec %>% update_role(date, new_role = "date")
+```
+
+In this case, the column is still available in the data and will be carried through the analysis (e.g. during resampling). This can be helpful since, if there is a poor prediction, this column may be helpful in diagnosing the problem. 
+
+The downside to chaining the role of the column is that it is needed at prediction-time. This can impact development tools (such as tuning functions) and would also be required to be in the data being predicted once the model is deployed. 
+
+If using a non-standard role for a column, it is highly recommended that you use the recipe within a workflow and specify `bake_dependent_roles` as described in [hardhat::default_recipe_blueprint()]. This will allow tuning and prediction functions to run without error.
```
juliasilge, line 38: If we're putting this in a workflow already in these docs, it would be very helpful (probably more helpful than the `bake_dependant_roles` demo) to show that you can `augment()` your fitted workflow using variables that aren't outcomes or predictors.

man/rmd/bake_dependent_roles.Rmd
juliasilge: I've spent a bit more time on this, and I do think we can do better than the `date` example for being clear about how to use these other roles.

juliasilge (changes requested):
I think we have this explanation in the right place now! 🙌

man/rmd/bake_dependent_roles.Rmd
topepo: This is exactly the application that has tried people up (us included). It seems most relevant. I can add another example though.

PR #1003, opened 2022-06-11
New family of spline steps

topepo:
These would eventually replace `step_bs()` and `step_ns()`. 

The splines2 package has a more rich set of options for splines. 

After initial review, I'll add steps based on `bSpline()`, `cSpline()`, `iSpline()`, `mSpline()`, and `bernsteinPoly()` (the last being made into `step_poly_bernstein()`. I'll also add soft deprecation notices for the original spline functions. 

This PR can come after version 1.0.0.


topepo:
A few of the spline functions have a `degree` argument. Can you add this as an argument and use the default value found in the underlying function? 

Also, can you rename `step_spline_basis()` to `step_spline_b()`? (since it is a "b-spline"). 

I'll do some more review and read up on the original paper on these.

topepo:
I think that we should default to `intercept = TRUE`

EmilHvitfeldt:
That should all be done now

PR #1006, opened 2022-06-14
document that step_spatialsign needs case weights at bake time

EmilHvitfeldt:


R/spatialsign.R
```diff
@@ -26,7 +26,19 @@
 #' When you [`tidy()`][tidy.recipe()] this step, a tibble with column
 #' `terms` (the columns that will be affected) is returned.
 #'
-#' @template case-weights-unsupervised
```
EmilHvitfeldt, line 29: The template isn't used because the extra text can't be added to the section created inside the template

PR #993, opened 2022-05-26
New role selectors

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/recipes/issues/942, close https://github.com/tidymodels/recipes/issues/669 and helps with https://github.com/tidymodels/recipes/issues/193.

It does this by expanding the role selectors we have available. 

without addressing the `*_predictors()` variants we used to have

- `all_numeric()`
- `all_nominal()`

This PR makes it so we have

- `all_numeric()`
    - `all_integer()`
    - `all_double()`
- `all_nominal()`
    - `all_ordered()`
    - `all_unordered()`
- `all_logical()`
- `all_date()`
- `all_datetime()`

The structure above is to indicate that `all_numeric()` captures both integers and doubles.

I made the changes to be backwards compatible. Even though `"nominal"` is no longer a type, it is still matched in change it was created using an old recipe. Furthermore it causes different results for recipes that did `has_type("numeric")` which IMO ins't too bad since they should have used `all_numeric()` anyways.

TODO:

- [ ] Fix tests
- [ ] update steps to work with changes, some steps uses testing [like this](https://github.com/tidymodels/recipes/blob/main/R/bin2factor.R#L96) which doesn't work anymore

PR #1007, opened 2022-06-16
Adds support for `tbl_lazy()` to `recipe()` 

edgararuiz:
This is the first PR that begins extending `recipes` into remote tables.  The idea is to setup the infrastructure in the main functions, as well as to centralize support for `tbl_lazy` objects.  These objects are used by `dplyr` on database connectivity packages, as well as `sparklyr`.  Highlights:

- It modifies the `recipe.data.frame()` function to enable it to accept non-data.frame objects.  Opted for this approach because of the current, and future, argument validation that happen inside this function.  The changes were minimal, so I figured it would be best to make them in this function, instead of copying the argument validation into another method
- Adds a `recipe.tbl_lazy()` method
- Renames `get_type()` to `recipes_translate_types()` and exports it.  This will allow other packages to pass custom type translation, such as `sparklyr`.
- Adds a `recipes_translate_types.tbl_lazy()` method. It will be the default translation for remote tables.  It collects the top 10 rows into memory, in order for the original `get_type()` code to determine types.  Because it is exported, this behavior can be customized based on the remote connection.  In the example below, I created a custom `tbl_spark` method for it. It uses the metadata about the field types already within the table object, so there is no need to collect the top 10.  I expect this function will be in `sparklyr` in a future release
-   Adds an un-exported `select_variables` method.  Its purpose is to add the ability to switch between using `tidyeval_eval_select()` for `data.frames` , and other methods to extract the selected variables if the object is not a `data.frame`. 
-  Added `select_variables.tbl_lazy()`.  Which emulates the output from the original `tidyeval::eval_select()`.  Today, it ignores the `allow_rename` argument. 

The resulting `recipe` object is identical as if was created using a local R `data.frame` with one exception.  The `template` item will not be a `data.frame` but a whatever the remote table is being used.  This is not an issue until a function such as `prep` is attempted.  This will be addressed in a subsequent PR. 

``` r
suppressPackageStartupMessages(library(sparklyr))
suppressPackageStartupMessages(library(recipes))
suppressPackageStartupMessages(library(DBI))

# ----- Example using a database connection

con <- dbConnect(RSQLite::SQLite(), dbname = ":memory:")

tbl_mtcars <- copy_to(con, mtcars)

remote_r <- recipe(tbl_mtcars) %>%
  step_select(mpg, am) %>%
  step_filter(am == 1)


remote_r
#> Recipe
#> 
#> Inputs:
#> 
#>   11 variables (no declared roles)
#> 
#> Operations:
#> 
#> Variables selected mpg, am
#> Row filtering using am == 1

# ----- Example using a Spark connection

sc <- spark_connect("local")

spark_mtcars <- copy_to(sc, mtcars)


# Created a custom type translation method using the new recipes_translate_types() function
recipes_translate_types.tbl_spark <- function(x) {
  type_map <- c(
    BooleanType = "logical",
    ByteType = "numeric",
    ShortType = "numeric",
    IntegerType = "numeric",
    FloatType = "numeric",
    DoubleType = "numeric",
    LongType = "numeric",
    StringType = "nominal",
    BinaryType = "nominal",
    TimestampType = "date",
    DateType = "date"
  )

  table_schema <- sdf_schema(x)

  purrr::map_dfr(table_schema, ~ {
    match_type <- names(type_map) == .x$type
    new_type <- type_map[match_type]
    tibble(variable = .x$name, type = new_type)
  })
}

spark_r <- recipe(spark_mtcars) %>%
  step_select(mpg, am) %>%
  step_filter(am == 1)

spark_r
#> Recipe
#> 
#> Inputs:
#> 
#>   11 variables (no declared roles)
#> 
#> Operations:
#> 
#> Variables selected mpg, am
#> Row filtering using am == 1

spark_disconnect(sc)

local_r <- recipe(mtcars) %>%
  step_select(mpg, am) %>%
  step_filter(am == 1)


local_r
#> Recipe
#> 
#> Inputs:
#> 
#>   11 variables (no declared roles)
#> 
#> Operations:
#> 
#> Variables selected mpg, am
#> Row filtering using am == 1
```

<sup>Created on 2022-06-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


R/misc.R
```diff
@@ -1,9 +1,15 @@
...
+#' data sources. It is expected that this method will be used by other packages
+#' to provide a translation of the data types.
+#' @param x Table object used to determine data types
+#' @export
+recipes_translate_types <- function(x) {
+  UseMethod("recipes_translate_types")
 }
 
 ## Buckets variables into discrete, mutally exclusive types
-get_types <- function(x) {
```
topepo, line 6: Take a look at #993 (specifically https://github.com/tidymodels/recipes/pull/993#discussion_r887321531). We might be expanding/refactoring how we do this part.

topepo (commented):
Just some initial thoughts on the code. I'll add Emil and Davis as reviewers too. 

It would be good to talk about this at the group meeting so we can discuss more broadly where things will be going.

R/misc.R
EmilHvitfeldt: This is the correct take. We should (once the dusts settles) make sure the whole `get_types()` system works well

EmilHvitfeldt (commented):
Other than the `get_types()` part which we have to work out separately I don't see anything in particular. Good PR!

R/selections.R
```diff
@@ -214,6 +214,38 @@ recipes_eval_select <- function(quos, data, info, ..., allow_rename = FALSE,
...
+
+select_variables.data.frame <- function(data, expr, allow_rename) {
+  tidyselect::eval_select(
+    expr = expr,
+    data = data,
+    allow_rename = allow_rename
+  )
+}
+
+select_variables.tbl_lazy <- function(data, expr, allow_rename) {
```
DavisVaughan, line 229: Maybe a comment about `allow_rename` purposefully being ignored for now

R/selections.R
```diff
@@ -195,7 +195,7 @@ recipes_eval_select <- function(quos, data, info, ..., allow_rename = FALSE,
   # Return names not positions, as these names are
   # used for both the training and test set and their positions
   # may have changed. If renaming is allowed, add the new names.
-  out <- names(data)[sel]
+  out <- extract_names(data)[sel]
```
DavisVaughan, line 198: I think `data` is always a data frame for us, so it is probably fine to simplify this and just use `colnames()` rather than `names()` and avoid adding a new S3 generic

DavisVaughan (commented):
I think my main overarching question is:

This doesn't work at all for recipes specified with a formula method right? Like:

```r
recipe(mpg ~ ., mtcars)
```

If `mtcars` is a remote table.

Do things like `update_role()` work? Like `recipe(mtcars) %>% update_role(mpg, new_role = "outcome")`

PR #1013, opened 2022-06-29
added another column to the summary method

topepo:


topepo:
An example: 

``` r
suppressPackageStartupMessages(library(recipes))

df <- tibble::tibble(y = c(1, 2, 3), x = c(4, 5, 6), var = c("a", "b", "c"))

rec_dummy <-
  recipe(y ~ ., df) %>%
  step_dummy(var, keep_original_cols = TRUE) %>%
  update_role(var, new_role = "id") %>%
  update_role_requirements("id", bake = FALSE) %>%
  step_center(x) %>%
  prep()

summary(rec_dummy, original = TRUE)
#> # A tibble: 3 × 5
#>   variable type    role      source   `required to bake`
#>   <chr>    <chr>   <chr>     <chr>    <lgl>             
#> 1 x        numeric predictor original TRUE              
#> 2 var      nominal id        original FALSE             
#> 3 y        numeric outcome   original FALSE
```

<sup>Created on 2022-06-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/recipe.R
```diff
@@ -730,7 +730,9 @@ print.recipe <- function(x, form_width = 30, ...) {
 #' @param ... further arguments passed to or from other methods (not currently
 #'   used).
 #' @return A tibble with columns `variable`, `type`, `role`,
-#'   and `source`.
+#'   and `source`. When `original = TRUE`, an additional column is included
+#'   named `required to bake` (based on the results of
```
DavisVaughan, line 734: Maybe `required_to_bake`? I try to avoid spaces in names, but if the `summary()` method is purely for interactive exploration, maybe it is okay

DavisVaughan (approved):
LGTM

PR #1014, opened 2022-06-30
New data checks

topepo:
@simonpcouch added a bunch of checks for cases when a user says they don't need a particular role at bake-time but actually does:

```
library(tidymodels)
tidymodels_prefer()

data(Chicago)

Chicago <- Chicago %>% select(ridership, date, one_of(stations))

n <- nrow(Chicago)
Chicago_train <- Chicago %>% slice(1:(n - 14))
Chicago_test  <- Chicago %>% slice((n - 13):n)

needs_bake <-
  recipe(ridership ~ ., data = Chicago_train) %>%
  # Create date features
  step_date(date) %>%
  step_holiday(date) %>%
  # Change date to be an id column instead of a predictor
  update_role(date, new_role = "id") %>%
  # Create dummy variables from factor columns
  step_dummy(all_nominal()) %>%
  prep()

# Davis's check_role_requirements() picks these up
bake(needs_bake, Chicago_test %>% select(-date))
#> Error in `bake()`:
#> ! The following required columns are missing from `new_data`: "date".
#> ℹ These columns have one of the following roles, which are required at `bake()` time: "id".
#> ℹ If these roles are not required at `bake()` time, use `update_role_requirements(role = "your_role", bake = FALSE)`.

# Now we say that we don't need 'date' even though we do. check_new_data()
# catches this case
still_needs_bake <-
  needs_bake %>%
  update_role_requirements("id", bake = FALSE) %>%
  prep()

bake(still_needs_bake, Chicago_test %>% select(-date))
#> Error in `step_date()`:
#> ! The following required column is missing from `new_data` in step
#>   'date_aLcii': date.
```


simonpcouch:
One note for posterity:

This PR leaves out checks for a few bake methods. Most (6) of them are column removal steps (search `(length(object$removals) > 0)` to see them), as it seems fine to me not to error when a column doesn’t exist when it only needs to be there to be removed.😶‍🌫️ Also leaves out `*_at` steps.

Will be adjusting the test cases in a sec to trigger only the non-standard role error.

PR #1044, opened 2022-10-06
Fix failing step_holiday() test on r-devel

EmilHvitfeldt:
This PR fixes the test that now fails, was first discovered https://github.com/tidymodels/recipes/actions/runs/3192498090/jobs/5210042479#step:6:234 and happened because of a change in r-devel https://developer.r-project.org/blosxom.cgi/R-devel/2022/10/05#n2022-10-05

### DIFF
--- a/tests/testthat/test_holiday.R
+++ b/tests/testthat/test_holiday.R
@@ -89,10 +89,8 @@ test_that("works with no missing values - Date class", {
 })
 
 test_that("POSIXct class", {
-  test_data$day <- as.POSIXct(test_data$day)
-  exp_dates$date <- as.POSIXct(exp_dates$date)
-  lubridate::tz(test_data$day) <- ""
-  lubridate::tz(exp_dates$date) <- ""
+  test_data$day <- lubridate::as_datetime(test_data$day, tz = "UTC")
+  exp_dates$date <- lubridate::as_datetime(exp_dates$date, tz = "UTC")
 
   holiday_rec <- recipe(~day, test_data) %>%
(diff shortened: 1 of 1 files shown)


tests/testthat/test_holiday.R
```diff
@@ -110,9 +110,16 @@ test_that("POSIXct class", {
     holiday_ind$day[is_equal_1(holiday_ind$day_Easter)],
     exp_dates$date[exp_dates$holiday == "Easter"]
   )
```
DavisVaughan, line 112: The real problem is the `as.POSIXct()` converters. I think you could insulate yourself from R-devel a little more by using lubridate converters.

i.e. use this at the top of the test instead

```r
test_data$day <- lubridate::as_datetime(test_data$day, tz = "UTC")
exp_dates$date <- lubridate::as_datetime(exp_dates$date, tz = "UTC")
```

and this in the test:

```r
lubridate::as_datetime(NA, tz = "UTC")
```

Those converter functions still aren't perfect but I think they should work here

EmilHvitfeldt: Done! Thank you!

tests/testthat/test_holiday.R
```diff
@@ -112,7 +112,7 @@ test_that("POSIXct class", {
...
   expect_equal(
     holiday_ind$day[is.na(test_data$day)],
-    as.POSIXct(NA, tz = NULL)
+    as.POSIXct(NA, tz = "UTC")
```
DavisVaughan, line 115: I don't trust `as.POSIXct()` here either, `lubridate::as_datetime(NA, tz = "UTC")`

PR #1051, opened 2022-10-27
Default to using clock locales inside step_date()

### LINKED ISSUE #1048: Reconsider handling of `locale` for `step_date()`
In #1000 `step_date()` was changed so the locale is recorded at `prep()` and then enforced at `bake()`. It turns out this presents problems for deploying a trained model to a different infrastructure. I noticed this when training a model on macOS (where I have `en_US.UTF-8`) and the trying to deploy on Linux. You can see [how the problem plays out here](https://colorado.rstudio.com/rsc/content/7ec7707d-064b-4343-8390-f87f0ab15335/test-locale.html).

That workflow (training locally on macOS, deploying on Linux) is pretty common but there are apparently more complex problems when Windows is involved.

The `locale` is only used for names of days of the week and months of the year, so is it possible we don't need the full locale system? Can we use [clock](https://clock.r-lib.org/reference/index.html#locale) instead?

EmilHvitfeldt:
This PR aims to close #1048. I switches to use `<clock_labels>` internally by default in `step_date()`.

The changes are backwards compatible, e.i. setting `locale = Sys.getlocale("LC_TIME")` still works as before.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,6 +21,7 @@ Depends:
     R (>= 3.4)
 Imports: 
     cli,
+    clock (>= 0.6.1),
     ellipsis,
     generics (>= 0.1.2),
     glue,
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 * Developer focused `.get_data_types()` generic has been added to designate types of columns. Exported for use in extension packages that deal with types not supported in recipes directly. (#993)
 
+* The `step_date()` function now defaults to using the clock package to format day-of-week and month labels. (#1048)
+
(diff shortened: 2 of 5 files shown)


EmilHvitfeldt:
The code in this step is starting to look kinda scary. I would suggest that once this PR is merged I do some follow-up PRs that clean up the insides to look like `step_time()`

juliasilge:
This looks great to me, as far as I can tell. It works for my use case. 👍  

The folks who maintain our Colorado server actually did install the other encoding on the server so I don't have any way to make a _failing_ example now, but this PR does work for me.

Issue #1005, opened 2022-06-14
POSIXct input causes step_ns and step_bs to fail at check_type()


jkennel:
### `check_type()` errors for POSIXct input in `step_ns()` and `step_bs()`

The splines package provides functions `ns()` and `bs()` which can handle POSIXct data. `step_ns()` and `step_bs()` fail because `check_type()` throws an error. This is relevant to modelling time series data and related to [#994](https://github.com/tidymodels/recipes/issues/994#issue-1249898279). See example below.

``` r
library(recipes)
#> Loading required package: dplyr
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
#> 
#> Attaching package: 'recipes'
#> The following object is masked from 'package:stats':
#> 
#>     step
library(splines)

# Generate data
set.seed(1)
dat <- data.frame(y = rnorm(100), 
                  x = seq.POSIXt(as.POSIXct('2022-06-14'),
                                 as.POSIXct('2022-06-15'),
                                 length.out = 100))
dat$x_num <- as.numeric(dat$x)  # numeric column


# These two calls work
head(splines::ns(dat$x, df = 5))
#>                 1 2           3          4           5
#> [1,] 0.000000e+00 0  0.00000000 0.00000000  0.00000000
#> [2,] 2.147104e-05 0 -0.01279448 0.03838344 -0.02558896
#> [3,] 1.717684e-04 0 -0.02552363 0.07657090 -0.05104727
#> [4,] 5.797182e-04 0 -0.03812213 0.11436640 -0.07624426
#> [5,] 1.374147e-03 0 -0.05052465 0.15157395 -0.10104930
#> [6,] 2.683881e-03 0 -0.06266586 0.18799758 -0.12533172
head(splines::bs(dat$x, df = 5))
#>               1           2            3 4 5
#> [1,] 0.00000000 0.000000000 0.000000e+00 0 0
#> [2,] 0.08682556 0.001351903 4.637746e-06 0 0
#> [3,] 0.16567883 0.005305581 3.710197e-05 0 0
#> [4,] 0.23685199 0.011707989 1.252191e-04 0 0
#> [5,] 0.30063723 0.020406081 2.968157e-04 0 0
#> [6,] 0.35732671 0.031246812 5.797182e-04 0 0

# Fails at check_type()
[...]

Issue #994, opened 2022-05-26
update check_type() to be more flexible 


EmilHvitfeldt:
Right now `check_type()` can test if variables are nominal or numeric. with the advent of https://github.com/tidymodels/recipes/pull/993 it would be nice to have a little more control since there is still quite a lot of manual checking in non-numeric/nominal steps

Issue #193, opened 2018-09-05
Helpers for boolean/logical columns


ClaytonJY:
There's a weird friction with modeling-in-R and logical columns; I'd argue that logical encoding is much more informative than 1/0 for most cases, but we can't trust modeling functions to convert those to 1/0 as needed. This means I often prepare a dataset to include logical columns (e.g. `is_foobar`) where appropriate, but then need to do something like `mutate_if(is.logical, as.integer)` just before moving into the `recipes` part of a modeling workflow.

Would it make sense to allow `recipes` to have helpers for logical columns?

I can imagine a few additions that would be helpful:
1. add a selector like `all_logical`, `all_boolean`, or even `all_bin`/`all_binary` (which might also grab 1/0 int cols and other 2-valued columns, though perhaps this is dangerous)
2. allow `step_bin2factor` to work with `logical` columns. This is often what I want to do, possibly followed by something like `step_dummy`, but I can't currently do that without making it an integer first, which can't be done in `recipes` AFAICT. Depending on interpretation of `bin`, I could also imagine this step handling _anything_ with only 2 values; strings, ints, doubles, logicals, etc.
3. add steps like `step_lgl2int`, `step_lgl2factor` (would be superceded by above)

My gut says that being overly broad in interpretation of "binary" is probably dangerous, and also untidy; best to stick with the type-safety encouraged by `purrr` and `rlang`? I think this also suggests `bin2factor` is not a great name; `int2factor` would be both more informative, and could be more broadly useful (would _have_ to give levels when >2, perhaps).

I suspect there might be other good ways to encode logical columns, but I think the core of the problem is allowing easy selection of them, and allowing conversion to 1/0 integers.

If this is a reasonable proposal I'm happy to do the implementation once a scope is determined.

reisner:
@ClaytonJY I agree this would be useful in general. 

For my particular use, I only had one logical column, so I was able to add this to my recipe:
```
step_mutate(logical_column = as.numeric(logical_column))
```

vspinu:
Why is there no `step_mutate_if`?

DavisVaughan:
Ideally you'd be able to use `step_mutate()` with `across(where(is.Date), fn)` but this bug keeps you from being able to do that https://github.com/tidymodels/recipes/issues/759

`mutate_if()` and `mutate_at()` are superseded in favor of `across()`, so I'd prefer not to add `step_mutate_if()` to recipes, and to mark `step_mutate_at()` as superseded

[...]

PR #1064, opened 2022-12-07
Add `types` argument to check_type()

EmilHvitfeldt:
This PR adds a new argument to `check_type()`. This way you can specify the required columns by `.get_data_types` types instead of using the more limited `quant`.

This PR should have no negative changes to extension packages as this new argument isn't the default and can be used on a opt-in basis.

Revdepcheck ran clear! (we needed revdepcheck because of new steps have started to use check_type()`)

This should close https://github.com/tidymodels/recipes/issues/1005 and close https://github.com/tidymodels/recipes/issues/994 and close https://github.com/tidymodels/recipes/issues/193


topepo (approved):
Look good. You might want to slack the vetiver people about `.get_data_types()`; it might help them resolve tidymodels issues the future. They'll know our "rules" for data types.

PR #1065, opened 2022-12-07
No more mod call args

EmilHvitfeldt:
We used to use `mod_call_args()`, but now we can use `rlang::call_modify()`. To close https://github.com/tidymodels/recipes/issues/266


hfrick (approved):
Nice! 🤓

Issue #953, opened 2022-04-14
Properly supersede `juice()`


EmilHvitfeldt:
Documentation wise we have stated that the function is superseded, but we need to properly link to https://lifecycle.r-lib.org/articles/stages.html#superseded with the badge.

PR #1066, opened 2022-12-07
properly Supersede juice

EmilHvitfeldt:
This PR finishes the superseding of `juice()` by

- Add supersede badge
- No longer use `juice()` in examples
- No longer use `juice()` in non-juice related tests

to close https://github.com/tidymodels/recipes/issues/953


hfrick (approved):
Looks good! 👍 

We could add one more reference to the new interface in the docs for `juice()` (`R/recipe.R` l809) and maybe we can avoid using juice as a verb in `R/update.R` l40. Maybe say bake or maybe just "to see what happened"?

PR #1069, opened 2022-12-12
Indicate which step caused the error

### LINKED ISSUE #420: Specify which step produces error

EmilHvitfeldt:
To close #420. Errors in `prep()` and `bake()` will now indicate which step caused the error. This is a pretty minimal code change, with many snapshot updates

# Before

``` r
library(recipes)

data("ames", package = "modeldata")

recipe(~., data = ames) |>
  step_novel(Neighborhood, new_level = "Gilbert") |>
  prep()
#> Error in `prep()`:
#> ! Columns already contain the new level: Neighborhood
```

# This PR

``` r
library(recipes)

data("ames", package = "modeldata")

recipe(~., data = ames) |>
  step_novel(Neighborhood, new_level = "Gilbert") |>
  prep()
#> Error in `step_novel()`:
#> ! Problem while executing step.
#> ! Columns already contain the new level: Neighborhood
```


simonpcouch (commented):
Love this thought! I think referencing the step-level call will be super helpful for users.

Given some error message:

```
Error in `step_YeoJohnson()`:
! Problem while executing step.
Caused by error in `FUN()`:
! Missing values are not allowed for the YJ transformation. See `na_rm` option
```

It feels to me like the `! Problem while executing step` is redundant with `Error in step_YeoJohnson():` and the `Caused by error in FUN():` almost always references an internal call rather than something the user wrote, as in a message constructed similarly for e.g. dplyr. The longer message also means that `conditionMessage()` output, condensed to one line, will almost surely be truncated by line width before the actual body of the error message, e.g.

```
Error in `step_YeoJohnson()`: ! Problem while executing step. Caused by error in...
```

How would you feel about instead just raising the step-level `call` and the condition message? e.g.

```
Error in `step_YeoJohnson()`:
! Missing values are not allowed for the YJ transformation. See `na_rm` option
```

which would be collapsed to

```
Error in `step_YeoJohnson()`: ! Missing values are not allowed for the YJ...
```

by `conditionMessage`.

tests/testthat/_snaps/R4.2/selections.md
```diff
@@ -3,7 +3,7 @@
     Code
       recipes_eval_select(quos = quos(I(beds:sqft)), data = Sacramento, info = info_sac)
     Condition
-      Error in `recipes_eval_select()`:
+      Error:
```
EmilHvitfeldt, line 6: This one is worse because we are testing the internal function by itself. This is fine as it will also be used within a function

tests/testthat/_snaps/filter.md
```diff
@@ -3,7 +3,8 @@
     Code
       prep(rec_1, training = iris %>% slice(1:75))
     Condition
-      Error in `dplyr::filter()`:
+      Error in `step_filter()`:
+      Caused by error in `dplyr::filter()`:
```
EmilHvitfeldt, line 7: This together with all the other {dplyr} steps give weird errors. This are still better than before. I will try to tackle these together in a different PR latter

tests/testthat/_snaps/logit.md
```diff
@@ -3,7 +3,8 @@
     Code
       prep(rec, training = ex_dat, verbose = FALSE)
     Condition
-      Error in `binomial()$linkfun()`:
+      Error in `step_logit()`:
+      Caused by error in `binomial()$linkfun()`:
```
EmilHvitfeldt, line 7: This will require a more extensive error handling, will tackle in future PR

tests/testthat/_snaps/selections.md
```diff
@@ -3,7 +3,7 @@
     Code
       recipes_eval_select(quos = quos(log(beds)), data = Sacramento, info = info_sac)
     Condition
-      Error in `recipes_eval_select()`:
+      Error:
```
EmilHvitfeldt, line 6: This is also a test for a internal function. So this is fine

tests/testthat/_snaps/discretized.md
```diff
@@ -74,15 +74,17 @@
     Condition
       Warning:
       Note that the options `prefix` and `labels` will be applied to all variables
-      Error in `cut.default()`:
+      Error in `step_discretize()`:
+      Caused by error in `cut.default()`:
```
simonpcouch, line 78: Is this a spot where we'd want to see `prep` in place of `cut.default`?

tests/testthat/_snaps/discretized.md
```diff
@@ -74,15 +74,17 @@
...
       ! lengths of 'breaks' and 'labels' differ
 
 # bad args
 
     Code
       recipe(~., data = ex_tr) %>% step_discretize(x1, num_breaks = 1) %>% prep()
     Condition
-      Error in `recipes::discretize()`:
+      Error in `step_discretize()`:
+      Caused by error in `recipes::discretize()`:
```
simonpcouch, line 87: Same goes for `prep` here.

simonpcouch (approved):
Awesome. I think this PR is a huge improvement to recipes' error messages!

tests/testthat/_snaps/discretized.md
EmilHvitfeldt: Same here

PR #1070, opened 2022-12-15
Fixes for dev dplyr

DavisVaughan:
This PR makes your package compatible with the next version of dplyr:

- `rename_with()` now checks that the result of the function is the right length. `paste0()` has bad recycling behavior when there are length 0 inputs so we typically use a `vec_paste0()` helper for this instead
- `slice()` with no inputs now drops all rows. We were testing this but i dont think we should test this edge case dplyr behavior so i just removed the test
- I also fixed a dev tidyr issue where we now assert that tidyselection in `drop_na()` shouldn't rename (because that isn't meaningful). And I used the more appropriate `all_of()` selector.

We plan to submit dplyr 1.1.0 on January 27th.

This should be compatible with both dev and CRAN dplyr. It would help us out if you could go ahead and send a patch version of your package to CRAN ahead of time! Thanks!

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -861,3 +861,8 @@ recipes_error_context <- function(expr, step_name) {
     }
   )
 }
+
+vec_paste0 <- function(..., collapse = NULL) {
+  args <- vctrs::vec_recycle_common(...)
+  rlang::inject(paste0(!!!args, collapse = collapse))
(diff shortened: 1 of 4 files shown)


EmilHvitfeldt:
Thank you!!

PR #1086, opened 2023-02-14
Update signed log description

zecojls:
The function is correct, but the description swapped log with abs for signed TRUE. I came here to make sure that log was being performed on abs values.


topepo:
Thanks for the contribution!

PR #1087, opened 2023-02-14
register `tune_args()` and `tunable()` methods unconditionally

simonpcouch:
Similar rationale and implementation as https://github.com/tidymodels/workflows/pull/193. :)


EmilHvitfeldt (approved):

R/tune_args.R
```diff
@@ -1,6 +1,9 @@
+#' @importFrom generics tune_args
+#' @export
+generics::tune_args
```
hfrick, line 3: this is already in `R/reexports.R` so we can remove it here. maybe there is a better place for it but for now I'd keep it consistent with the rest of the package.

hfrick: I left comments about not exporting the generic for `tune_args()` in the other PRs, but for recipes it's a little different. Since it was already exported, I wouldn't change that here, this note is more for reference.
Full paper trail: the reason this is already exported is that the generic for `tunable()` was already exported: https://github.com/tidymodels/recipes/pull/830#discussion_r728377566

hfrick (approved):
Just one comment about code duplication 👍

PR #1088, opened 2023-02-14
remove `tune_check()` test helper

simonpcouch:
Similar rationale and implementation as https://github.com/tidymodels/parsnip/pull/870. :)

PR #1091, opened 2023-02-16
add `conditionMessage()` method for `recipes_error`s

### LINKED ISSUE #1080: add `conditionMessage` method for `recipes_error`s
I’m very much appreciating recipes pointing out which step errors occurred in following \#1069. We unfortunately don’t get the benefits of those changes in tune (and possibly other packages that need to catch and rethrow errors from recipes, see tidymodels/tune#599), as dispatch falls back to the `rlang_error` method for `conditionMessage()`:

``` r
library(recipes)

data("ames", package = "modeldata")

err <- try(
  recipe(~., data = ames) |>
  step_novel(Neighborhood, new_level = "Gilbert") |>
  prep()
)
#> Error in step_novel() : 
#> ! Columns already contain the new level: Neighborhood

err_cnd <- attr(err, "condition")

class(err_cnd)
#> [1] "recipes_error_step" "recipes_error"      "rlang_error"       
#> [4] "error"              "condition"

conditionMessage(err_cnd)
#> [1] "\nCaused by error in `prep()`:\n! Columns already contain the new level: Neighborhood"
```

Could we add a `conditionMessage` method that precedes the current `conditionMessage()` output with “Error in `step_*`”?

<sup>Created on 2023-01-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

simonpcouch:
Closes #1080.

Just lightly wraps the `conditionMessage` method for `rlang_error`s to set the `prefix` argument to `TRUE` so that we can see which step caused the error in `conditionMessage()` output. tune's erroring machinery will pick up on this change "for free." :)


simonpcouch:
Same failure to install dials on Ubuntu as in #1088.

EmilHvitfeldt:
GHA failures have been fixed

topepo (approved):

Issue #1089, opened 2023-02-15
remove copy of `s3_register()`


hfrick:
there is a `R/s3_register.R` file which contains a `s3_register()` function. this might be a copy of the vctrs function. we don't seem to use it anywhere so it'd be good to remove it

PR #1106, opened 2023-03-14
remove unused s3_register()

EmilHvitfeldt:
To close https://github.com/tidymodels/recipes/issues/1089

`s3_register()` was never used so this will be safe


simonpcouch (approved):
Ah, nice!

PR #1109, opened 2023-03-21
Avoid depending on the specifics of "object not found" base R errors

EmilHvitfeldt:
This PR modifies some functions so that we don't rely on base R errors, and deletes a couple of tests of the same nature where the errors pop up in dplyr functions

### DIFF
--- a/R/roles.R
+++ b/R/roles.R
@@ -307,6 +307,9 @@ update_role <- function(recipe, ..., new_role = "predictor", old_role = NULL) {
 #' @rdname roles
 #' @export
 remove_role <- function(recipe, ..., old_role) {
+  if (rlang::is_missing(old_role)) {
+    rlang::abort('argument "old_role" is missing, with no default.')
+  }
   single_chr(old_role, "old_")
 
   terms <- quos(...)
--- a/R/selections.R
+++ b/R/selections.R
@@ -172,6 +172,10 @@ recipes_eval_select <- function(quos, data, info, ..., allow_rename = FALSE,
                                 check_case_weights = TRUE, call = caller_env()) {
   ellipsis::check_dots_empty()
 
+  if (rlang::is_missing(quos)) {
+    rlang::abort('argument "quos" is missing, with no default.')
+  }
+
   # Maintain ordering between `data` column names and `info$variable` so
   # `eval_select()` and recipes selectors return compatible positions
   data_info <- tibble(variable = names(data))
--- a/tests/testthat/_snaps/filter.md
+++ b/tests/testthat/_snaps/filter.md
@@ -1,14 +1,3 @@
-# quasiquotation
-
-    Code
-      prep(rec_1, training = iris %>% slice(1:75))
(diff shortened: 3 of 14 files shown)


R/selections.R
```diff
@@ -172,6 +172,10 @@ recipes_eval_select <- function(quos, data, info, ..., allow_rename = FALSE,
                                 check_case_weights = TRUE, call = caller_env()) {
   ellipsis::check_dots_empty()
 
+  if (rlang::is_missing(quos)) {
+    rlang::abort('argument "quos" is missing, with no default.')
```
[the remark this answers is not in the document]
simonpcouch: Looks like the `call` is already passed into this function!

R/roles.R
```diff
@@ -307,6 +307,9 @@ update_role <- function(recipe, ..., new_role = "predictor", old_role = NULL) {
 #' @rdname roles
 #' @export
 remove_role <- function(recipe, ..., old_role) {
+  if (rlang::is_missing(old_role)) {
+    rlang::abort('argument "old_role" is missing, with no default.')
```
simonpcouch, line 311:
```suggestion
    rlang::abort('argument "old_role" is missing, with no default.', call = call)
```

R/roles.R
```diff
@@ -307,6 +307,9 @@ update_role <- function(recipe, ..., new_role = "predictor", old_role = NULL) {
 #' @rdname roles
 #' @export
 remove_role <- function(recipe, ..., old_role) {
```
simonpcouch, line 309:
```suggestion
remove_role <- function(recipe, ..., old_role, call = caller_env()) {
```

I'm not exactly sure where all this is called, but one level higher up in the call stack seems more likely to be familiar to a user.

simonpcouch (approved):
I'm with it!

R/selections.R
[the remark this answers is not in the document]
EmilHvitfeldt: I'm going to label this as an `.internal = TRUE` because it should only pop up if this function is used incorrectly by the maintainer

R/roles.R
EmilHvitfeldt: This function is already the highest level it needs to be as it is user-facing. 

``` r
library(recipes)
data("biomass", package = "modeldata")

recipe(x = biomass) |>
  update_role(sample, new_role = "role1") |>
  remove_role(sample)
#> Error in `remove_role()`:
#> ! argument "old_role" is missing, with no default.
#> Backtrace:
#>     ▆
#>  1. └─recipes::remove_role(...)
#>  2.   └─rlang::abort("argument \"old_role\" is missing, with no default.") at recipes/R/roles.R:311:4
```

<sup>Created on 2023-03-22 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

R/roles.R
EmilHvitfeldt: Answered in other suggestion

PR #1111, opened 2023-03-22
remove unused filter_terms()

EmilHvitfeldt:
What the title says. This function is no longer used, and isn't exported


simonpcouch (approved):
Woot woot!

PR #1113, opened 2023-03-22
don't use ellipse_check()

EmilHvitfeldt:
ellipse_check() is deprecated and it appears like we forgot to stop using it in `step_nnmf_sparse()`

https://github.com/tidymodels/recipes/blob/f8f02691c977ee0039122661d4e3d48b0dbd29ea/R/misc.R#L275-L279


simonpcouch (approved):
I'm with it! Related to #813.

PR #1114, opened 2023-03-22
delete unused terms.recipe()

EmilHvitfeldt:
What the title says. This function is no longer used, and isn't exported


simonpcouch (approved):
Dropped a `browser()` in here and ran `test()` and didn't see any usages. On board!🦜

PR #1115, opened 2023-03-29
transition from `map_dfr()` in `tune_args.recipe()`

simonpcouch:
Just because this function gets so much traffic, the deprecation machinery for `map_dfr()` ended up causing some slowdowns. :)


EmilHvitfeldt (approved):
Looking great!

PR #1117, opened 2023-03-29
use hardhat::recompose()

EmilHvitfeldt:
Related to https://github.com/tidymodels/hardhat/pull/220 and https://github.com/tidymodels/hardhat/pull/231

Short short, `convert_matrix()` was in both {recipes} and {hardhat}, now with this change, it is only in {hardhat}


simonpcouch (approved):
Awesome! Feels great to delete so many lines.

I think this happened in the meantime between you requesting review and me reviewing, but with hardhat 1.3.0 on CRAN, looks like we can remove the Remote and bump the minimum version? RSPM allowing.😆

PR #1118, opened 2023-03-30
pass allow_rename directly to tidyselect::eval_select()

EmilHvitfeldt:
As listed in the file, we needed to wait for https://github.com/r-lib/tidyselect/issues/221 to be fixed.

This was included in the 1.2.0 release, which we are already depending on so we should be good there.


simonpcouch (approved):
Golden! Goodbye, technical debt🌟

Issue #1097, opened 2023-02-22
Make backwards compatible change to step_range default correctly


EmilHvitfeldt:
As noted in https://github.com/tidymodels/recipes/pull/1093#issuecomment-1437369355, we need to make sure that if clipping isn't set, that we default to clip.

PR #1119, opened 2023-03-30
perform clipping if clipping is missing in step_range()

EmilHvitfeldt:
To close https://github.com/tidymodels/recipes/issues/1097


simonpcouch (approved):
Yup, looks great. Thanks for leaving such a thorough paper trail in your issue/PR descriptions. :)

PR #1126, opened 2023-04-17
align file names with step names

EmilHvitfeldt:
This PR is the first in a long series of refactoring. 

It:
- renames `R/` files to match name of steps
- renames `tests/` files to match name of `R/` for steps
- splits up `test-center_scale_norm.R` into 3 separate files


hfrick (approved):
Nice quality of life improvement for living the usethis workflow ✨

PR #1127, opened 2023-04-20
Only use `[[` for subsetting single columns

EmilHvitfeldt:
This is another refactor/standardize PR

I noticed that we are using every method under the sun to pull out a single column from a data.frame. This PR aims to only use `[[` for consistency.

Not that it matters hugely, but this also provides the smallest of speedups

``` r
col_name <- "vs"
bench::mark(
  mtcars[, col_name, drop = TRUE],
  getElement(mtcars, col_name),
  mtcars[[col_name]]
)
#> # A tibble: 3 × 6
#>   expression                           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 mtcars[, col_name, drop = TRUE]   2.62µs   3.03µs   303949.   13.28KB     30.4
#> 2 getElement(mtcars, col_name)      2.05µs   2.34µs   403596.    4.88KB     40.4
#> 3 mtcars[[col_name]]                1.72µs   1.97µs   464780.        0B     46.5
```

### DIFF
--- a/R/shuffle.R
+++ b/R/shuffle.R
@@ -90,10 +90,11 @@ bake.step_shuffle <- function(object, new_data, ...) {
 
   if (length(object$columns) > 0) {
     for (i in seq_along(object$columns)) {
-      new_data[, object$columns[i]] <-
-        sample(getElement(new_data, object$columns[i]))
+      new_data[[object$columns[i]]] <-
+        sample(new_data[[object$columns[i]]])
     }
   }
+
   new_data
 }
 
--- a/R/BoxCox.R
+++ b/R/BoxCox.R
@@ -135,7 +135,7 @@ bake.step_BoxCox <- function(object, new_data, ...) {
   check_new_data(param, object, new_data)
 
   for (i in seq_along(object$lambdas)) {
-    new_data[, param[i]] <- bc_trans(getElement(new_data, param[i]), lambda = object$lambdas[i])
+    new_data[, param[i]] <- bc_trans(new_data[[param[i]]], lambda = object$lambdas[i])
   }
 
   new_data
--- a/R/BoxCox.R
+++ b/R/BoxCox.R
@@ -135,7 +135,10 @@ bake.step_BoxCox <- function(object, new_data, ...) {
   check_new_data(param, object, new_data)
 
   for (i in seq_along(object$lambdas)) {
-    new_data[, param[i]] <- bc_trans(new_data[[param[i]]], lambda = object$lambdas[i])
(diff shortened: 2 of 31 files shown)


R/shuffle.R
```diff
@@ -90,10 +90,11 @@ bake.step_shuffle <- function(object, new_data, ...) {
 
   if (length(object$columns) > 0) {
     for (i in seq_along(object$columns)) {
-      new_data[, object$columns[i]] <-
-        sample(getElement(new_data, object$columns[i]))
+      new_data[[object$columns[i]]] <-
+        sample(new_data[[object$columns[i]]])
     }
```
EmilHvitfeldt, line 95: Some of these changes doesn't give super nice readability right now as I find `]]]` a little hard to read.

I will address these in a future PR where I will refactor many of the loops for readability (and speed)

```r
# Old
if (length(object$columns) > 0) {
  for (i in seq_along(object$columns)) {
      new_data[[object$columns[i]]] <-
        sample(new_data[[object$columns[i]]])
    }
}

# New
col_names <- object$columns
for (col_name in col_names) {
  new_data[[col_name]] <- sample(new_data[[col_name]])
}
```

simonpcouch (approved):
TIL about `getElement()`👀

For the GitHub sleuths:

There are a couple changes that aren't inset `<-` and don't set `drop` and transition `x[, i]` to `x[[i]]` that made me a bit nervous. For data frames, there's no change here with default `drop = TRUE`, but for tibbles:

``` r
# `x[, 2]` is not equal to `x[[2]]`
mtcars_tbl <- tibble::as_tibble(mtcars)

mtcars_tbl[, 2]
#> # A tibble: 32 × 1
#>      cyl
#>    <dbl>
#>  1     6
#>  2     6
#>  3     4
#>  4     6
#>  5     8
#>  6     6
#>  7     8
#>  8     4
#>  9     4
#> 10     6
#> # ℹ 22 more rows

mtcars_tbl[[2]]
#>  [1] 6 6 4 6 8 6 8 4 4 6 6 8 8 8 8 8 8 4 4 4 4 8 8 8 8 4 4 4 8 6 8 4
```

<sup>Created on 2023-04-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

In these cases, results were passed to `complete.cases()`, `is.na()`, or inset again, all of which are handled consistently between methods. So, thumbs up. :)

Issue #1130, opened 2023-04-25
Major test reorganization issue


EmilHvitfeldt:
I'm going to be doing some test organizations. This will include:

- moving tests around
- get standardized naming on common tests
- fill in gaps of missing tests

This issue will serve as an anchor, in what will properly be many PRs.

The first PR that would have fallen under this project is

- https://github.com/tidymodels/recipes/pull/1126

EmilHvitfeldt:
This has been done now!

PR #1131, opened 2023-04-25
add infrastructure section to all step test files

EmilHvitfeldt:
This PR is part of https://github.com/tidymodels/recipes/issues/1130.

This PR seems like it is doing a lot, but it is just moving tests around. Looking at it commit by commit is very helpful.

I'm defining infrastructure tests as tests that should be used on all steps, or all applicable tests. All steps should test that they print correctly, and all steps that uses `check_name()` should test that they are used. These are infrastructure tests.

Tests under the infrastructure section should be in alphabetical order, for no other reason than to avoid confusion when new tests are added

- Each `step_*()` and `check_*()` file gets an `# Infrastructure ----------` section in their test file.
- renamed `test-factors2string.R` to match name of step
- split `mutate()` and `rename()` tests into 2 each to the `_at()` variants have their own files.

Moved over tests that should be called for each step


EmilHvitfeldt:
Missing test

# Empty printing

```r
[1] "tests/testthat/test-geodist.R"                                                                   
[2] "tests/testthat/test-interact.R"   
[3] "tests/testthat/test-intercept.R"  
[4] "tests/testthat/test-mutate.R"     
[5] "tests/testthat/test-nnmf_sparse.R"
[6] "tests/testthat/test-sample.R"     
[7] "tests/testthat/test-select.R"    
```

EmilHvitfeldt:
Incorrect tests

# empty selection tidy method works

The following steps don't return 0-row tibbles

- poly_bernstein
- spline_b
- spline_convex
- spline_monotone
- spline_natural
- spline_nonnegative
- profile

Issue #1134, opened 2023-04-26
Add missing infrastructure tests


EmilHvitfeldt:
# Empty printing

```r
c(
"tests/testthat/test-geodist.R",                                                    
"tests/testthat/test-interact.R",
"tests/testthat/test-intercept.R",
"tests/testthat/test-mutate.R", # one test called "no input"
"tests/testthat/test-nnmf_sparse.R",
"tests/testthat/test-sample.R",
"tests/testthat/test-select.R"
)
```

# printing

```r
c(
"tests/testthat/test-newvalues.R",
"tests/testthat/test-nnmf_sparse.R"                              
)
```

# empty selection tidy method works

```r
c("
tests/testthat/test-geodist.R", 
"tests/testthat/test-interact.R",                                  
"tests/testthat/test-intercept.R", 
"tests/testthat/test-lag.R", 
"tests/testthat/test-mutate.R", 
"tests/testthat/test-nnmf_sparse.R", 
"tests/testthat/test-sample.R", 
"tests/testthat/test-select.R"
)
```
# empty selection prep/bake is a no-op

```r
c("tests/testthat/test-arrange.R", 
"tests/testthat/test-classdist.R",                                 
"tests/testthat/test-count.R", 
"tests/testthat/test-depth.R", 
"tests/testthat/test-geodist.R",
 "tests/testthat/test-interact.R", 
"tests/testthat/test-intercept.R", 
"tests/testthat/test-mutate.R", 
"tests/testthat/test-nnmf_sparse.R",
"tests/testthat/test-regex.R", 
"tests/testthat/test-sample.R", 
"tests/testthat/test-select.R"
)
```
# bake method errors when needed non-standard role columns are missing

Some of these might be false negatives. E.i. I need to think about if all steps needs this test

```r
c(
"tests/testthat/test-arrange.R",
 "tests/testthat/test-class.R",                                     
"tests/testthat/test-colcheck.R", 
"tests/testthat/test-corr.R", 
"tests/testthat/test-dummy_multi_choice.R", 
"tests/testthat/test-filter.R", 
"tests/testthat/test-filter_missing.R", 
"tests/testthat/test-intercept.R", 
"tests/testthat/test-lincomb.R", 
"tests/testthat/test-missing.R", 
"tests/testthat/test-mutate.R", 
"tests/testthat/test-mutate_at.R", 
"tests/testthat/test-naomit.R", 
"tests/testthat/test-newvalues.R", 
[...]

PR #1138, opened 2023-04-29
let step_interact() work with empty selection

EmilHvitfeldt:
This bug came up when I started working on https://github.com/tidymodels/recipes/issues/1134

## PR

``` r
library(recipes)

recipe(mpg ~ ., mtcars) |>
  step_interact() |>
  prep() |>
  bake(new_data = NULL)
#> # A tibble: 32 × 11
#>      cyl  disp    hp  drat    wt  qsec    vs    am  gear  carb   mpg
#>    <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl> <dbl>
#>  1     6  160    110  3.9   2.62  16.5     0     1     4     4  21  
#>  2     6  160    110  3.9   2.88  17.0     0     1     4     4  21  
#>  3     4  108     93  3.85  2.32  18.6     1     1     4     1  22.8
#>  4     6  258    110  3.08  3.22  19.4     1     0     3     1  21.4
#>  5     8  360    175  3.15  3.44  17.0     0     0     3     2  18.7
#>  6     6  225    105  2.76  3.46  20.2     1     0     3     1  18.1
#>  7     8  360    245  3.21  3.57  15.8     0     0     3     4  14.3
#>  8     4  147.    62  3.69  3.19  20       1     0     4     2  24.4
#>  9     4  141.    95  3.92  3.15  22.9     1     0     4     2  22.8
#> 10     6  168.   123  3.92  3.44  18.3     1     0     4     4  19.2
#> # ℹ 22 more rows
```

## Main

``` r
library(recipes)

recipe(mpg ~ ., mtcars) |>
  step_interact() |>
  prep() |>
  bake(new_data = NULL)
#> Error in step_interact_new(terms = terms, trained = trained, role = role, : argument "terms" is missing, with no default
```

### DIFF
--- a/R/interact.R
+++ b/R/interact.R
@@ -93,7 +93,7 @@ step_interact <-
     add_step(
       recipe,
       step_interact_new(
-        terms = terms,
+        terms = enquos(terms),
         trained = trained,
         role = role,
         objects = objects,
@@ -124,8 +124,27 @@ step_interact_new <-
 ## one large set of collected terms.
 #' @export
 prep.step_interact <- function(x, training, info = NULL, ...) {
+
+  # Empty selection
+  if (identical(x$terms[[1]], quo())) {
+    return(
+        step_interact_new(
+        terms = x$terms,
+        role = x$role,
+        trained = TRUE,
+        objects = x$objects,
+        sep = x$sep,
+        skip = x$skip,
+        id = x$id
+      )
+    )
+  }
+
   # Identify any selectors that are involved in the interaction
   # formula
+  tmp_terms <- as.formula(rlang::as_label(x$terms[[1]]))
+  environment(tmp_terms) <- environment(x$terms[[1]])
+  x$terms <- tmp_terms
 
   form_sel <- find_selectors(x$terms)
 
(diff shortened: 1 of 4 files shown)


R/interact.R
```diff
@@ -93,7 +93,7 @@ step_interact <-
     add_step(
       recipe,
       step_interact_new(
-        terms = terms,
+        terms = enquos(terms),
```
EmilHvitfeldt, line 96: This is the original cause of the problem, since `terms` is a required argument, it wouldn't let it be empty. turning it into quosures like the rest of the steps treats selectors seems like the right choice

R/interact.R
```diff
@@ -124,8 +124,27 @@ step_interact_new <-
...
+        id = x$id
+      )
+    )
+  }
+
   # Identify any selectors that are involved in the interaction
   # formula
+  tmp_terms <- as.formula(rlang::as_label(x$terms[[1]]))
+  environment(tmp_terms) <- environment(x$terms[[1]])
+  x$terms <- tmp_terms
```
EmilHvitfeldt, line 147: These lines of code switches the list of quosures back into the formula the rest of the function expects

EmilHvitfeldt:
@topepo I'm requesting review from you since this feels like a bitter change.

The only scenario where this would have been sketchy would be if an un-prepped recipe was saved, from an earlier version and loaded in. These changes are backwards compatible

topepo (approved):

PR #1139, opened 2023-05-01
Standardize columns param documentation

EmilHvitfeldt:
Turned out that the `columns` argument in the different steps did the same thing, and was just documented in a bunch of different wants. This PR fixes that and uses the updated `step_pca()` version


hfrick (approved):
🧹 ✨

PR #1142, opened 2023-05-10
Add remaining infrastructure tests for empty selections

EmilHvitfeldt:
Part of overall project: https://github.com/tidymodels/recipes/issues/1130
dealing with part of https://github.com/tidymodels/recipes/issues/1134 by adding tests for

- Empty printing
- printing
- empty selection prep/bake is a no-op

Doing the other 2 types in a different PR for ease of review.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 # recipes (development version)
 
-* Fixed bug where `step_interact()` didn't work with empty selection.
+* Fixed bugs where `step_classdist()`, `step_count()`, `step_depth()`,  `step_geodist()`,  `step_interact()`, `step_nnmf_sparse()`, and  `step_regex()` didn't work with empty selection.
+
+* `step_classdist()`, `step_count()` and `step_depth()` no longer returns a column with all `NA`s with empty selections.
+
+* `step_regex()` no longer returns a column with all 0s with empty selections.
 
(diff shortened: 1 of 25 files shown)


NEWS.md
hfrick, line 1: Given that different steps did different things for "empty selection", it'd be nice to add what they do now (which is "nothing").

tests/testthat/_snaps/geodist.md
```diff
@@ -162,6 +162,41 @@
       Error in `step_geodist()`:
       ! `is_lat_lon` should be a single logical value.
 
+# empty printing
+
+    Code
+      rec
```
hfrick, line 168: What do you think about adding the definition of `rec` to the snapshots? Then we could see that it's testing the right thing without referring to the `test-*.R` files.

EmilHvitfeldt: it is not a bad idea. 

I think I like to stay with the current design, as all the steps follow this formula with a couple of small variations.

```r
test_that("empty printing", {
  rec <- recipe(mpg ~ ., mtcars)
  rec <- name_of_step_here(rec)

  expect_snapshot(rec)

  rec <- prep(rec, mtcars)

  expect_snapshot(rec)
})

```

Issue #1133, opened 2023-04-26
Fix incorrect tests for empty selection tidy methods


EmilHvitfeldt:
# empty selection tidy method works

The following steps don't return 0-row tibbles

- poly_bernstein
- spline_b
- spline_convex
- spline_monotone
- spline_natural
- spline_nonnegative
- profile

_Originally posted by @EmilHvitfeldt in https://github.com/tidymodels/recipes/issues/1131#issuecomment-1522486578_

PR #1143, opened 2023-05-11
make sure all steps return zero row tibble with empty selection

EmilHvitfeldt:
To close https://github.com/tidymodels/recipes/issues/1133.
related to https://github.com/tidymodels/recipes/issues/1130

a couple of steps didn't return zero-row tibbles with empty selections. This PR fixes that


hfrick (approved):
Looks good! 🧹 🎉

PR #1144, opened 2023-05-12
make sure all steps have `empty selection tidy method works` tests

EmilHvitfeldt:
What the title says. part of https://github.com/tidymodels/recipes/issues/1134 and https://github.com/tidymodels/recipes/issues/1130

### DIFF
--- a/R/sample.R
+++ b/R/sample.R
@@ -51,7 +51,8 @@
 #'
 #' bake(smaller_cars, new_data = NULL) %>% nrow()
 #' bake(smaller_cars, new_data = mtcars %>% slice(21:32)) %>% nrow()
-step_sample <- function(recipe, ...,
+step_sample <- function(recipe,
+                        ...,
                         role = NA,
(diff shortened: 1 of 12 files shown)


R/sample.R
```diff
@@ -170,9 +171,14 @@ print.step_sample <-
...
-  )
+  if (is.null(x$size)) {
+    res <- tibble(size = numeric(), replace = logical())
+  } else {
+    res <- tibble(
+      size = x$size,
+      replace = x$replace
+    )
+  }
+  res$id <- x$id
```
simonpcouch, line 182: It looks like this may have previously referenced `x$inputs` rather than `x$id`?

simonpcouch (approved):
Nice! Two small suggestions. :)

R/sample.R
EmilHvitfeldt: yes, and it shouldn't have. `$inputs` doesn't exists, and will return `NULL`

PR #1146, opened 2023-05-16
Use vec_cbind() instead of bind_cols()

EmilHvitfeldt:
Small speed improvement, little downside for the way we use it.

```r
library(vctrs)
library(dplyr)

mtcars1 < mtcars[, 1:4]
mtcars2 <- mtcars[, -(1:4)]

bench::mark(
  bind_cols(mtcars1, mtcars2),
  vec_cbind(mtcars1, mtcars2)
)
#> # A tibble: 2 × 6
#>   expression                       min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>                  <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 bind_cols(mtcars1, mtcars2)   51.9µs   54.5µs    17939.  111.13KB     38.7
#> 2 vec_cbind(mtcars1, mtcars2)   18.8µs     20µs    48725.    1.18KB     39.0

```


simonpcouch (approved):
Awesome. :)

I had initially worried that `vec_cbind()` has slightly different logic about tibble return type (via `vec_ptype2()`) than `bind_cols()` does. A nice bit of insurance we have here, though:

https://github.com/tidymodels/recipes/blob/97e45ff6b21ae889f21f1a98c427f9ac8a5030e3/R/recipe.R#L644-L656

We'd be seeing that error if any inputs inside of these `bake()` methods were not already tibbles. Huzzah!🚀

PR #1149, opened 2023-05-27
add remove_original_cols() function

EmilHvitfeldt:
I noticed we used the same pattern all over the package. This PR provides a small function to make our lives easier.

This change should make it

- called outside the loop
- small speed increase
- should remove a lot of inconsistency


simonpcouch (commented):
One small comment!

The only thing that struck me about this PR was the lack of `isFALSE()`/`isTRUE()` when handling `keep_original_cols` input, as there are no checks that `keep_original_cols` is a logical in `step_*()`:

``` r
library(recipes)

mtcars$cyl <- as.factor(mtcars$cyl)
  
recipe(mpg ~ ., data = mtcars) %>%
  step_dummy(cyl, id = "", keep_original_cols = "boop")
#> 
#> ── Recipe ──────────────────────────────────────────────────────────────────────
#> 
#> ── Inputs
#> Number of variables by role
#> outcome:    1
#> predictor: 10
#> 
#> ── Operations
#> • Dummy variables from: cyl
```

Before this PR, these errors surfaced like:

``` r
recipe(mpg ~ ., data = mtcars) %>%
  step_dummy(cyl, id = "", keep_original_cols = "boop") %>%
  prep()
#> Error in `step_dummy()`:
#> Caused by error in `!object$preserve`:
#> ! invalid argument type
```

With this PR, this now looks like:

``` r
recipe(mpg ~ ., data = mtcars) %>%
  step_dummy(cyl, id = "", keep_original_cols = "boop") %>%
  prep()
#> Error in `step_dummy()`:
#> Caused by error in `!keep_original_cols`:
#> ! invalid argument type
```

<sup>Created on 2023-06-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

...which is definitely better! Anticipating that a bigger push on input checking will come somewhere down the line, thumbs up from me. :)

PR #1153, opened 2023-06-10
add recipes_remove_cols()

EmilHvitfeldt:
This developer function adds a little more standardization. I decided it would be helpful when writing `remove_original_cols()`.

I didn't apply this function in `step_bs()` and `step_ns()` because they should be updated to use `remove_original_cols()` which this PR is set to do https://github.com/tidymodels/recipes/pull/1154


R/misc.R
```diff
@@ -860,7 +860,37 @@ vec_paste0 <- function(..., collapse = NULL) {
...
-    new_data <- new_data[, !(colnames(new_data) %in% col_names), drop = FALSE]
+    new_data <- recipes_remove_cols(new_data, object, col_names)
+  }
+  new_data
+}
+
+#' Removes columns if options apply
+#'
+#' This helper function should whenever columns should be removed using
+#' character vectors.
```
hfrick, line 871:
```suggestion
#' This helper function removes columns based on character vectors.
```

hfrick (approved):
Looks good!

Issue #661, opened 2021-03-09
Use tidyverse style to document `step_*()` functions


hfrick:
The tidyverse style guide suggests that function names should be followed by `()`. The first line of the documentation for (most) `step*()` functions reads
```
`step_*` creates a *specification* of a recipe step
```
so the `step_*` should get some `()` added.

https://style.tidyverse.org/documentation.html#r-code

PR #1159, opened 2023-06-16
update descriptions for all step functions

EmilHvitfeldt:
To close https://github.com/tidymodels/recipes/issues/661

I originally found the following unfortunate description: "step_unorder creates a specification of a recipe step that will transform the data.". So I decided to take a look at everything while I was at it

### DIFF
--- a/R/BoxCox.R
+++ b/R/BoxCox.R
@@ -1,8 +1,7 @@
 #' Box-Cox Transformation for Non-Negative Data
 #'
-#' `step_BoxCox` creates a *specification* of a recipe
-#'  step that will transform data using a simple Box-Cox
-#'  transformation.
+#' `step_BoxCox()` creates a *specification* of a recipe step that will transform
+#' data using a Box-Cox transformation.
 #'
 #' @inheritParams step_center
 #' @param lambdas A numeric vector of transformation values. This
--- a/R/YeoJohnson.R
+++ b/R/YeoJohnson.R
@@ -1,8 +1,7 @@
 #' Yeo-Johnson Transformation
 #'
-#' `step_YeoJohnson` creates a *specification* of a
-#'  recipe step that will transform data using a simple Yeo-Johnson
-#'  transformation.
+#' `step_YeoJohnson()` creates a *specification* of a recipe step that will
+#' transform data using a Yeo-Johnson transformation.
 #'
 #' @inheritParams step_center
 #' @param lambdas A numeric vector of transformation values. This
--- a/R/cut.R
+++ b/R/cut.R
@@ -1,7 +1,7 @@
 #' Cut a numeric variable into a factor
 #'
 #' `step_cut()` creates a *specification* of a recipe step that cuts a numeric
-#'  variable into a factor based on provided boundary values
+#' variable into a factor based on provided boundary values
 #'
 #' @inheritParams step_center
 #' @param breaks A numeric vector with at least one cut point.
--- a/R/cut.R
+++ b/R/cut.R
@@ -1,7 +1,7 @@
 #' Cut a numeric variable into a factor
 #'
 #' `step_cut()` creates a *specification* of a recipe step that cuts a numeric
-#' variable into a factor based on provided boundary values
+#' variable into a factor based on provided boundary values.
 #'
 #' @inheritParams step_center
 #' @param breaks A numeric vector with at least one cut point.
--- a/R/depth.R
+++ b/R/depth.R
@@ -1,7 +1,7 @@
 #' Data Depths
 #'
 #' `step_depth()` creates a *specification* of a recipe step that will convert
-#' numeric data into measurement of *data depth*. This is done for each value of
+#' numeric data into a measurement of *data depth*. This is done for each value of
 #' a categorical class variable.
 #'
 #' @inheritParams step_pca
--- a/R/depth.R
+++ b/R/depth.R
@@ -1,9 +1,8 @@
 #' Data Depths
 #'
-#' `step_depth` creates a *specification* of a recipe
-#'  step that will convert numeric data into measurement of
-#'  *data depth*. This is done for each value of a categorical
-#'  class variable.
+#' `step_depth()` creates a *specification* of a recipe step that will convert
+#' numeric data into measurement of *data depth*. This is done for each value of
+#' a categorical class variable.
 #'
 #' @inheritParams step_pca
 #' @inheritParams step_center
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -1,9 +1,8 @@
 #' Create traditional dummy variables
 #'
-#' `step_dummy()` creates a *specification* of a recipe
-#'  step that will convert nominal data (e.g. character or factors)
-#'  into one or more numeric binary model terms for the levels of
-#'  the original data.
+#' `step_dummy()` creates a *specification* of a recipe step that will convert
+#' nominal data (e.g. character or factors) into one or more numeric binary
+#' model terms for the levels of the original data.
 #'
 #' @inheritParams step_pca
 #' @inheritParams step_center
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -1,8 +1,8 @@
 #' Create traditional dummy variables
 #'
 #' `step_dummy()` creates a *specification* of a recipe step that will convert
-#' nominal data (e.g. character or factors) into one or more numeric binary
-#' model terms for the levels of the original data.
+#' nominal data (e.g. characters or factors) into one or more numeric binary
+#' model terms corresponding to the levels of the original data.
 #'
 #' @inheritParams step_pca
 #' @inheritParams step_center
--- a/R/dummy_extract.R
+++ b/R/dummy_extract.R
@@ -1,8 +1,8 @@
 #' Extract patterns from nominal data
 #'
-#' `step_dummy_extract()` creates a *specification* of a recipe
-#'  step that will convert nominal data (e.g. character or factors)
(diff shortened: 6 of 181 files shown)


R/BoxCox.R
```diff
@@ -1,8 +1,7 @@
 #' Box-Cox Transformation for Non-Negative Data
 #'
-#' `step_BoxCox` creates a *specification* of a recipe
-#'  step that will transform data using a simple Box-Cox
-#'  transformation.
+#' `step_BoxCox()` creates a *specification* of a recipe step that will transform
+#' data using a Box-Cox transformation.
```
simonpcouch, line 4: Googled this and it does seem to be the case, but just confirming: "simple" here is a filler word that doesn't clarify meaning and is not a technical term describing the procedure more exactly?

R/YeoJohnson.R
```diff
@@ -1,8 +1,7 @@
 #' Yeo-Johnson Transformation
 #'
-#' `step_YeoJohnson` creates a *specification* of a
-#'  recipe step that will transform data using a simple Yeo-Johnson
-#'  transformation.
+#' `step_YeoJohnson()` creates a *specification* of a recipe step that will
+#' transform data using a Yeo-Johnson transformation.
```
simonpcouch, line 4: Same comment re: deletion of "simple" here.

R/cut.R
```diff
@@ -1,7 +1,7 @@
 #' Cut a numeric variable into a factor
 #'
 #' `step_cut()` creates a *specification* of a recipe step that cuts a numeric
-#'  variable into a factor based on provided boundary values
+#' variable into a factor based on provided boundary values
```
simonpcouch, line 4:
```suggestion
#' variable into a factor based on provided boundary values.
```

R/depth.R
```diff
@@ -1,9 +1,8 @@
 #' Data Depths
 #'
-#' `step_depth` creates a *specification* of a recipe
-#'  step that will convert numeric data into measurement of
-#'  *data depth*. This is done for each value of a categorical
-#'  class variable.
+#' `step_depth()` creates a *specification* of a recipe step that will convert
+#' numeric data into measurement of *data depth*. This is done for each value of
```
simonpcouch, line 4:
```suggestion
#' numeric data into a measurement of *data depth*. This is done for each value of
```

I think so? Or "measurement" -> "measurements"

R/dummy.R
```diff
@@ -1,9 +1,8 @@
 #' Create traditional dummy variables
 #'
-#' `step_dummy()` creates a *specification* of a recipe
-#'  step that will convert nominal data (e.g. character or factors)
-#'  into one or more numeric binary model terms for the levels of
-#'  the original data.
+#' `step_dummy()` creates a *specification* of a recipe step that will convert
+#' nominal data (e.g. character or factors) into one or more numeric binary
```
simonpcouch, line 4:
```suggestion
#' nominal data (e.g. characters or factors) into one or more numeric binary
```

R/dummy.R
```diff
@@ -1,9 +1,8 @@
...
+#' model terms for the levels of the original data.
```
simonpcouch, line 5:
```suggestion
#' model terms corresponding to the levels of the original data.
```

R/dummy_extract.R
```diff
@@ -1,8 +1,8 @@
 #' Extract patterns from nominal data
 #'
-#' `step_dummy_extract()` creates a *specification* of a recipe
-#'  step that will convert nominal data (e.g. character or factors)
-#'  into one or more integer model terms for the extracted levels.
+#' `step_dummy_extract()` creates a *specification* of a recipe step that will
+#' convert nominal data (e.g. character or factors) into one or more integer
```
simonpcouch, line 4:
```suggestion
#' convert nominal data (e.g. characters or factors) into one or more integer
```

R/dummy_multi_choice.R
```diff
@@ -1,9 +1,8 @@
 #' Handle levels in multiple predictors together
 #'
-#' `step_dummy_multi_choice()` creates a *specification* of a recipe
-#'  step that will convert multiple nominal data (e.g. character or factors)
-#'  into one or more numeric binary model terms for the levels of
-#'  the original data.
+#' `step_dummy_multi_choice()` creates a *specification* of a recipe step that
+#' will convert multiple nominal data (e.g. character or factors) into one or
```
simonpcouch, line 4:
```suggestion
#' will convert multiple nominal data (e.g. characters or factors) into one or
```

simonpcouch: What does the "multiple" refer to here? Columns?

R/factor2string.R
```diff
@@ -1,6 +1,7 @@
 #' Convert Factors to Strings
 #'
-#' `step_factor2string` will convert one or more factor vectors to strings.
+#' `step_factor2string()`  creates a *specification* of a recipe step that will
```
simonpcouch, line 3:
```suggestion
#' `step_factor2string()` creates a *specification* of a recipe step that will
```

R/harmonic.R
```diff
@@ -1,7 +1,7 @@
 #' Add sin and cos terms for harmonic analysis
 #'
-#' `step_harmonic` creates a *specification* of a recipe step that
-#'   will add sin and cos terms for harmonic analysis.
+#' `step_harmonic()` creates a *specification* of a recipe step that will add
+#' sin and cos terms for harmonic analysis.
```
simonpcouch, line 4:
```suggestion
#' `sin()` and `cos()` terms for harmonic analysis.
```

Or `$math font$`? I'm not sure.

R/intercept.R
```diff
@@ -1,11 +1,10 @@
...
 #'
-#' `step_intercept` creates a *specification* of a recipe step that
-#'   will add an intercept or constant term in the first column of a data
-#'   matrix. `step_intercept` has defaults to *predictor* role so
-#'   that it is by default called in the bake step. Be careful to avoid
-#'   unintentional transformations when calling steps with
-#'   `all_predictors`.
+#' `step_intercept()` creates a *specification* of a recipe step that will add
+#' an intercept or constant term in the first column of a data matrix.
+#' `step_intercept()` has defaults to *predictor* role so that it is by default
```
simonpcouch, line 5:
```suggestion
#' `step_intercept()` defaults to *predictor* role so that it is by default
```

R/intercept.R
```diff
@@ -1,11 +1,10 @@
...
+#' called in the bake step. Be careful to avoid unintentional transformations
```
simonpcouch, line 6:
```suggestion
#' only called in the bake step. Be careful to avoid unintentional transformations
```

Is this what this sentence means?

R/isomap.R
```diff
@@ -1,8 +1,8 @@
 #' Isomap Embedding
 #'
-#' `step_isomap` creates a *specification* of a recipe
-#'  step that will convert numeric data into one or more new
-#'  dimensions.
+#' `step_isomap()` creates a *specification* of a recipe step that uses
+#' multidimensional scaling to convert numeric data into one or more new
+#' dimensions using
```
simonpcouch, line 5: I believe this one wasn't completed.

R/lincomb.R
```diff
@@ -1,8 +1,8 @@
 #' Linear Combination Filter
 #'
-#' `step_lincomb` creates a *specification* of a recipe
-#'  step that will potentially remove numeric variables that have
-#'  linear combinations between them.
+#' `step_lincomb()` creates a *specification* of a recipe step that will
+#' potentially remove numeric variables that have linear combinations between
+#' them.
```
simonpcouch, line 5: How about 

> "that are linear combinations of others."

?

I'm not familiar with the currently used verbiage for linear combinations.

R/relevel.R
```diff
@@ -1,10 +1,9 @@
 #' Relevel factors to a desired level
 #'
-#' `step_relevel` creates a *specification* of a recipe
-#'  step that will reorder the provided factor columns so that
-#'  the level specified by ref_level is first. This is useful
-#'  for contr.treatment contrasts which take the first level as the
-#'  reference.
+#' `step_relevel()` creates a *specification* of a recipe step that will reorder
+#' the provided factor columns so that the level specified by `ref_level` is
+#' first. This is useful for `contr.treatment()` contrasts which take the first
```
simonpcouch, line 5:
```suggestion
#' first. This is useful for [contr.treatment()] contrasts which take the first
```

simonpcouch (changes requested):
Nice! These are definitely an improvement. :)

R/BoxCox.R
EmilHvitfeldt: simple is filler in this case! :)

R/dummy_multi_choice.R
EmilHvitfeldt: columns

R/isomap.R
EmilHvitfeldt: fixed!

R/lincomb.R
EmilHvitfeldt: So, after reading up on this, it technically removed Linear Dependency. Linear combinations are when 2 or more vectors are combined in `v1 * a1 + v2 * a2 + ... + vn * an` with scalar `a`s.

Linear dependency is what happens when a linear combination is equal to 0.

In the following, we have that `c = a + b` so the variables have a linear combination between them. could also be written as `0 = a + b - c`.
```r
a <- c(0, 1, 1, 0)
b <- c(1, 0, 0, 1)
c <- c(1, 1, 1, 1)
```

PR #1163, opened 2023-06-20
start adding developer function section

EmilHvitfeldt:
There are quite a lot of functions that are used to create steps and to interact with recipes. 

This PR aims to add a documentation section that gives an overview of all of them


R/developer.R
```diff
@@ -0,0 +1,162 @@
+#' @name developer_functions
+#'
+#' @title Developer functions for creating recipes steps
+#'
+#' @description
+#'
+#' This page provides a comprehensive list of the exported functions for
+#' creating recipes steps, and how to use them.
```
simonpcouch, line 8:
```suggestion
#' creating recipes steps and guidance on how to use them.
```

R/developer.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' @details
+#'
+#' # Creating steps
+#'
+#' [add_step()] and [add_check()] are required when creating a new step. It
+#' should be the other function call in your steps and should have the following
```
simonpcouch, line 15: "It should be the other function call" is ambiguous to me. Based on the shown code, maybe this could read "The output of `add_step()` should be the return value of all steps..."?

R/developer.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' [rand_id()] should be used in the arguments of `step_example()` to specify
+#' the argument, as we see in the above example.
+#'
+#' [recipes_pkg_check()] should be used in `step_example()` functions together
+#' with [required_pkgs()] to alert users that certain other packages are
+#' required. The standard way of using this function is the following format:
+#'
+#' ```r
+#' recipes_pkg_check(required_pkgs.step_example())
```
simonpcouch, line 46: Is it actually standard to write out the name of the method? If not, I'd write it using only the name of the `required_pkgs()` generic and note that it ought to dispatch to the `step_example` method.

R/developer.R
```diff
@@ -0,0 +1,162 @@
...
+#'     role = role,
+#'     trained = trained,
+#'     skip = skip,
+#'     id = id
+#'   )
+#' }
+#' ```
+#'
+#' [recipes_eval_select()] is used within `prep.step_*()` functions, and are
+#' used to turn `x$terms` object into a character vector of the selected
```
simonpcouch, line 66:
```suggestion
#' used to turn the `terms` object into a character vector of the selected
```

R/developer.R
```diff
@@ -0,0 +1,162 @@
...
+#'
+#' ```r
+#' col_names <- recipes_eval_select(x$terms, training, info)
+#' ```
+#'
+#' [check_type()] can be used within `prep.step_*()` functions to check that the
+#' variables passed in are the right types. We recommend that you use the
+#' `types` argument as it offers higher flexibility and it matches the types
+#' defined by [.get_data_types()]. When using `types` we find it better to be
+#' explicit, so writing `types = c("double", "integer")` instead of `types =
```
simonpcouch, line 79:
```suggestion
#' explicit, e.g. writing `types = c("double", "integer")` instead of `types =
```

R/developer.R
```diff
@@ -0,0 +1,162 @@
...
+#' "numeric"` as it produces cleaner error messages.
```
simonpcouch, line 80:
```suggestion
#' "numeric"`, as it produces cleaner error messages.
```

R/developer.R
```diff
@@ -0,0 +1,162 @@
...
+#' new columns to the data set. The function throws an error if the column names
+#' already exist in the data set. It should be called before adding the new
+#' columns to the data set.
+#'
+#' [get_keep_original_cols()] and [remove_original_cols()] are used within steps
+#' with the `keep_original_cols` argument. [get_keep_original_cols()] is used in
+#' `prep.step_*()` functions, for steps that were created before the
+#' `keep_original_cols` argument was added. And acts as a way to throw a warning
+#' that the user should regenerate the recipe. [remove_original_cols()] should
+#' be used in `bake.step_*()` functions to help remove the original cols. It is
```
simonpcouch, line 108:
```suggestion
#' be used in `bake.step_*()` functions to remove the original columns. It is
```

R/developer.R
```diff
@@ -0,0 +1,162 @@
...
+#' worth noting that [remove_original_cols()] can remove multiple columns at
+#' once and when possible should be put outside `for` loops.
+#'
+#' ```r
+#' new_data <- remove_original_cols(new_data, object, names_of_original_cols)
+#' ```
+#'
+#' [recipes_remove_cols()] should be used in `prep.step_*()` functions, and are
```
simonpcouch, line 116:
```suggestion
#' [recipes_remove_cols()] should be used in `prep.step_*()` functions, and is
```

R/developer.R
EmilHvitfeldt: This is actually how we use it! 😄 

https://github.com/tidymodels/recipes/blob/66a1542b861af752703eb5b4593e71235df05622/R/pls.R#L143

we have to do it this way since we call it before an `step_pls` object is created for `required_pkgs()` to dispatch off

R/developer.R
EmilHvitfeldt: fixed in [`fa88266` (#1163)](https://github.com/tidymodels/recipes/pull/1163/commits/fa8826634aed9ffb712174485ac8ee081ceb1f74)

R/developer.R
simonpcouch: Okay, gotcha! Thumbs up, then. :)

PR #1166, opened 2023-06-21
Unify tests for keep_original_cols arguments

EmilHvitfeldt:
This PR aims to make sure the `keep_original_cols` argument as adequate and consistent test coverage.

Appears that all steps were using `keep_original_cols` correctly. Most tests did not tests for one of the two options `keep_original_cols = FALSE` or `keep_original_cols = TRUE`. The tests do now.

Moreover, many tests are now run with slightly simpler code and less data.


simonpcouch (approved):
I dig it!⛏

PR #1156, opened 2023-06-13
Loop cleaning

EmilHvitfeldt:
This is quite a big PR.

it tries to give a more strieght forward unified approach to how to do things in `bake_step_*()` functions.

The gist of this, is that for many cases, we want to have the functions use the following structure

```r
for (col_name in col_names) {
  do something with  new_data[[col_name]]
}
```

It also has the benefit that the column names `col_names` are only extracted once, so we will gain a little speed and consistency that way.

Getting rid of `map()` inside `bake_step_*()` functions is also beneficial since they mess up the error handling/passing we are trying to use.

# Todo

The following steps needs these changes but for various reason didn't get done now. I'm planning on addressing these in separate PRs first.

- [x] `step_window()` https://github.com/tidymodels/recipes/pull/1172
    - is a 1 to many step, but doens’t act like it
- [x] `step_regex()` https://github.com/tidymodels/recipes/pull/1169
    - similar to step_dummy(), missing keep_orginal_cols()
- [x] `step_ratio()` https://github.com/tidymodels/recipes/pull/1173
    - doesn’t feel right
- [x] `step_profile()` https://github.com/tidymodels/recipes/pull/1177
    - also feels weird
- [x] `step_ns()` https://github.com/tidymodels/recipes/pull/1178
    - more work
- [x] `step_lag()` https://github.com/tidymodels/recipes/pull/1179
    - is a 1 to many step, but doesn’t act like it
- [x] `step_impute_linear()`, `step_impute_knn()`, `step_impute_bag()` https://github.com/tidymodels/recipes/pull/1184
    - missing lots of things. such as recipes_eval_select()
- [x] `step_depth()` https://github.com/tidymodels/recipes/pull/1180
    - missing check_new_data()
- [x] `step_count() ` https://github.com/tidymodels/recipes/pull/1169
    - is similar to step_regex()
- [x] `step_classdist()` https://github.com/tidymodels/recipes/pull/1183
    - similar to step_dummy() missing keep_orginal_cols()
- [x] `step_bs()` https://github.com/tidymodels/recipes/pull/1178
    - more work
- [x] `step_dummy()` https://github.com/tidymodels/recipes/pull/1182

## example benchmark

```r
old <- function(new_data, object) {
  if (length(object$columns) > 0) {
    for (i in seq_along(object$columns)) {
      new_data[, object$columns[i]] <-
        sample(getElement(new_data, object$columns[i]))
    }
  }
  new_data
}

new <- function(new_data, object) {
[...]

PR #1169, opened 2023-07-21
refactor step_regex() and step_count()

EmilHvitfeldt:

This PR does a small refactor to `step_regex()` and `step_count()` to bring them up to the same standard as #1156.

Further, the steps new use `check_name()`, and will thus error in `result` is set to a name that is in the data set. Previously it would silently overwrite.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # recipes (development version)
 
+* `step_regex()` and `step_count()` will now informatively error if name collision occurs.
+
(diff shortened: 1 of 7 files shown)


simonpcouch (approved):
This is great! 

A paper trail for future passersby: I was missing a bit of context while trying to understand how many of these lines transition `eval_select()` output `object$input` to `names(object$input)`. I initially had trouble wrapping my head around how that was fully safe (e.g. depending on `object$input` being `c(Species = "Species")`), and this seemed to suggest to me not:

https://github.com/tidymodels/recipes/blob/59486ec5cf0d345eebdc5e278a69e8883de3e232/R/selections.R#L204-L208

This is fine for both `step_count()` and `step_regex()`, though, because both leave `allow_rename = FALSE` as default in `recipes_eval_select()` and don't surface that argument to the user.🐛

PR #1177, opened 2023-07-31
loopify step_profile()

EmilHvitfeldt:

This PR does a small refactor to `step_profile()` to bring them up to the same standard as https://github.com/tidymodels/recipes/pull/1156.

This step is quite different that most other steps as it doesn't use the `new_data` in `bake()`, so many of the check functions we normally use shouldn't be done here.

### DIFF
--- a/R/profile.R
+++ b/R/profile.R
@@ -217,7 +217,7 @@ bake.step_profile <- function(object, new_data, ...) {
 
   # Keep the predictors in the same order
   keepers <- names(new_data)[names(new_data) %in% keepers]
-  new_data <- dplyr::select(new_data, !!keepers)
+  new_data <- new_data[keepers]
   new_data
 }
 
--- a/R/profile.R
+++ b/R/profile.R
@@ -200,17 +200,24 @@ prep.step_profile <- function(x, training, info = NULL, ...) {
 
 #' @export
 bake.step_profile <- function(object, new_data, ...) {
-  n <- length(object$profile[[1]])
-  new_data <- new_data[rep(1, n), ]
-  keepers <- c(names(object$columns), names(object$profile))
+  col_names <- names(object$columns)
+  profile_names <- names(object$profile)
+  keepers <- c(col_names, profile_names)
+
+  new_data <- list()
+
+  for (col_name in col_names) {
+    new_data[[col_name]] <- object$columns[[col_name]]
+  }
+
+  new_data[[names(object$profile)]] <- object$profile[[1]]
+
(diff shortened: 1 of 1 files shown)


R/profile.R
```diff
@@ -200,17 +200,24 @@ prep.step_profile <- function(x, training, info = NULL, ...) {
...
-  # Keep the predictors in the same order
-  keepers <- names(new_data)[names(new_data) %in% keepers]
-  new_data <- dplyr::select(new_data, !!keepers)
+  col_names <- names(object$columns)
+  profile_names <- names(object$profile)
+  keepers <- c(col_names, profile_names)
 
-  for (i in names(object$columns)) {
-    new_data[[i]] <- rep(object$columns[[i]], n)
+  new_data <- list()
```
EmilHvitfeldt, line 207: I found it easier to create a empty list, fill it, order it, tibblify it. Than to modify the existing tibble, order it making sure the lengths matched

R/profile.R
```diff
@@ -200,17 +200,24 @@ prep.step_profile <- function(x, training, info = NULL, ...) {
...
+
+  for (col_name in col_names) {
+    new_data[[col_name]] <- object$columns[[col_name]]
   }
+
   new_data[[names(object$profile)]] <- object$profile[[1]]
+
+  new_data <- vctrs::vec_recycle_common(!!!new_data)
```
EmilHvitfeldt, line 215: This function is so cool!

``` r
aaa <- list(1, 2, 4, 1:5)

vctrs::vec_recycle_common(!!!aaa)
#> [[1]]
#> [1] 1 1 1 1 1
#> 
#> [[2]]
#> [1] 2 2 2 2 2
#> 
#> [[3]]
#> [1] 4 4 4 4 4
#> 
#> [[4]]
#> [1] 1 2 3 4 5
```

R/profile.R
```diff
@@ -200,17 +200,24 @@ prep.step_profile <- function(x, training, info = NULL, ...) {
...
+  new_data <- tibble::new_tibble(new_data)
+
+  # Keep the predictors in the same order
+  keepers <- names(new_data)[names(new_data) %in% keepers]
+  new_data <- dplyr::select(new_data, !!keepers)
```
simonpcouch, line 220: Since `keepers` is a character vector, could we use

```suggestion
  new_data <- new_data[keepers]
```

to bypass tidyselect for the performance speedup?

simonpcouch (approved):
I dig it! One note on a pattern we may be able to use elsewhere in these PRs. :)

PR #1178, opened 2023-07-31
loopify `step_bs()` and `step_ns()`

EmilHvitfeldt:

This PR does a small refactor to `step_bs()` and `step_ns()` to bring them up to the same standard as https://github.com/tidymodels/recipes/pull/1156.

### DIFF
--- a/R/bs.R
+++ b/R/bs.R
@@ -173,10 +173,10 @@ bake.step_bs <- function(object, new_data, ...) {
 
   for (col_name in col_names) {
     new_values <- bs_predict(object$objects[[col_name]], new_data[[col_name]])
-    new_values <- as_tibble(new_values)
-    new_names <- paste(col_name, "bs", names0(ncol(new_values), ""), sep = "_")
 
+    new_names <- paste(col_name, "bs", names0(ncol(new_values), ""), sep = "_")
     colnames(new_values) <- new_names
+
     new_values <- check_name(new_values, new_data, object, new_names)
     new_data <- vctrs::vec_cbind(new_data, new_values)
   }
--- a/R/bs.R
+++ b/R/bs.R
@@ -178,7 +178,7 @@ bake.step_bs <- function(object, new_data, ...) {
 
     colnames(new_values) <- new_names
     new_values <- check_name(new_values, new_data, object, new_names)
-    new_data <- vec_cbind(new_data, new_values)
+    new_data <- vctrs::vec_cbind(new_data, new_values)
   }
 
   new_data <- remove_original_cols(new_data, object, col_names)
--- a/R/bs.R
+++ b/R/bs.R
@@ -168,29 +168,20 @@ prep.step_bs <- function(x, training, info = NULL, ...) {
 
 #' @export
 bake.step_bs <- function(object, new_data, ...) {
-  check_new_data(names(object$objects), object, new_data)
+  col_names <- names(object$objects)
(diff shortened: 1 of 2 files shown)


R/bs.R
```diff
@@ -168,29 +168,20 @@ prep.step_bs <- function(x, training, info = NULL, ...) {
...
+  check_new_data(col_names, object, new_data)
+
+  for (col_name in col_names) {
+    new_values <- bs_predict(object$objects[[col_name]], new_data[[col_name]])
+    new_values <- as_tibble(new_values)
+    new_names <- paste(col_name, "bs", names0(ncol(new_values), ""), sep = "_")
+
+    colnames(new_values) <- new_names
+    new_values <- check_name(new_values, new_data, object, new_names)
+    new_data <- vec_cbind(new_data, new_values)
```
topepo, line 181: We should namespace these to avoid issues with parallel processing.

topepo (changes requested):

R/bs.R
EmilHvitfeldt: Done!

R/bs.R
```diff
@@ -168,29 +168,20 @@ prep.step_bs <- function(x, training, info = NULL, ...) {
...
-      paste(orig_var, "bs", names0(new_cols[i], ""), sep = "_")
-    colnames(bs_values)[cols] <- new_names
-    strt <- max(cols) + 1
-    new_data <- remove_original_cols(new_data, object, orig_var)
+  col_names <- names(object$objects)
+  check_new_data(col_names, object, new_data)
+
+  for (col_name in col_names) {
+    new_values <- bs_predict(object$objects[[col_name]], new_data[[col_name]])
+    new_values <- as_tibble(new_values)
```
simonpcouch, line 176: I'm a bit hesitant about bringing `as_tibble()` inside of this loop, with an eye for performance. I tried pulling it out, as 

``` r
new_data <- as_tibble(new_data)
```

and it seemed to work fine. Any thoughts on this?

R/ns.R
```diff
@@ -159,27 +159,20 @@ prep.step_ns <- function(x, training, info = NULL, ...) {
...
-      paste(orig_var, "ns", names0(new_cols[i], ""), sep = "_")
-    colnames(ns_values)[cols] <- new_names
-    strt <- max(cols) + 1
-    new_data <- remove_original_cols(new_data, object, orig_var)
+  col_names <- names(object$objects)
+  check_new_data(col_names, object, new_data)
+
+  for (col_name in col_names) {
+    new_values <- bs_predict(object$objects[[col_name]], new_data[[col_name]])
+    new_values <- as_tibble(new_values)
```
simonpcouch, line 167: Same thought, and seemingly successful experiment, here.

simonpcouch (commented):
These look a _lot_ cleaner!

R/bs.R
EmilHvitfeldt: I guess this is fine, the reason I did it because I wasn't expecting `vec_cbind()` to work with matrix input, but it appears to work well.

``` r
vctrs::vec_cbind(
  tibble::tibble(x = 2), 
  matrix(1, dimnames = list(NULL, "y"))
)
#> # A tibble: 1 × 2
#>       x     y
#>   <dbl> <dbl>
#> 1     2     1
```

In that case we wouldn't need `as_tibble()` at all. Because is guaranteed  that `new_data` will be a tibble

PR #1179, opened 2023-07-31
loopify step_lag()

EmilHvitfeldt:

This PR does a small refactor to `step_lag()` to bring them up to the same standard as https://github.com/tidymodels/recipes/pull/1156.

### DIFF
--- a/R/lag.R
+++ b/R/lag.R
@@ -115,7 +115,7 @@ bake.step_lag <- function(object, new_data, ...) {
   check_new_data(col_names, object, new_data)
 
   for (col_name in col_names) {
-    new_values <- map(
+    new_values <- lapply(
       object$lag,
       function(x) dplyr::lag(new_data[[col_name]], x, default = object$default)
     )
--- a/R/lag.R
+++ b/R/lag.R
@@ -92,7 +92,7 @@ step_lag_new <-
 #' @export
 prep.step_lag <- function(x, training, info = NULL, ...) {
   if (!all(x$lag == as.integer(x$lag))) {
-    rlang::abort("step_lag requires 'lag' argument to be integer valued.")
+    rlang::abort("step_lag() requires 'lag' argument to be integer-valued.")
   }
 
   step_lag_new(
@@ -120,7 +120,7 @@ bake.step_lag <- function(object, new_data, ...) {
       function(x) dplyr::lag(new_data[[col_name]], x, default = object$default)
     )
 
-    new_names <- glue("{object$prefix}{object$lag}_{col_name}")
+    new_names <- glue::glue("{object$prefix}{object$lag}_{col_name}")
     names(new_values) <- new_names
 
     new_values <- tibble::new_tibble(new_values)
--- a/R/lag.R
+++ b/R/lag.R
@@ -91,6 +91,10 @@ step_lag_new <-
 
 #' @export
 prep.step_lag <- function(x, training, info = NULL, ...) {
+  if (!all(x$lag == as.integer(x$lag))) {
+    rlang::abort("step_lag requires 'lag' argument to be integer valued.")
+  }
+
   step_lag_new(
     terms = x$terms,
     role = x$role,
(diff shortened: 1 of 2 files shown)


R/lag.R
```diff
@@ -107,29 +111,24 @@ prep.step_lag <- function(x, training, info = NULL, ...) {
...
-  grid <- tidyr::expand_grid(col = object$columns, lag_val = object$lag)
-  calls <- purrr::map2(grid$col, grid$lag_val, make_call)
-  newname <- as.character(glue("{object$prefix}{grid$lag_val}_{grid$col}"))
-  calls <- check_name(calls, new_data, object, newname, TRUE)
+    new_names <- glue("{object$prefix}{object$lag}_{col_name}")
+    names(new_values) <- new_names
+
+    new_values <- tibble::new_tibble(new_values)
+    new_values <- check_name(new_values, new_data, object, new_names)
+    new_data <- vec_cbind(new_data, new_values)
```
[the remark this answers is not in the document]
EmilHvitfeldt: done!

R/lag.R
```diff
@@ -107,29 +111,24 @@ prep.step_lag <- function(x, training, info = NULL, ...) {
...
-
-  make_call <- function(col, lag_val) {
-    call2(
-      "lag",
-      x = sym(col),
-      n = lag_val,
-      default = object$default,
-      .ns = "dplyr"
+  for (col_name in col_names) {
+    new_values <- map(
```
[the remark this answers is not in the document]
EmilHvitfeldt: Done!

R/lag.R
```diff
@@ -107,29 +111,24 @@ prep.step_lag <- function(x, training, info = NULL, ...) {
...
+      object$lag,
+      function(x) dplyr::lag(new_data[[col_name]], x, default = object$default)
     )
-  }
 
-  grid <- tidyr::expand_grid(col = object$columns, lag_val = object$lag)
-  calls <- purrr::map2(grid$col, grid$lag_val, make_call)
-  newname <- as.character(glue("{object$prefix}{grid$lag_val}_{grid$col}"))
-  calls <- check_name(calls, new_data, object, newname, TRUE)
+    new_names <- glue("{object$prefix}{object$lag}_{col_name}")
```
simonpcouch, line 123:
```suggestion
    new_names <- glue::glue("{object$prefix}{object$lag}_{col_name}")
```

R/lag.R
```diff
@@ -91,6 +91,10 @@ step_lag_new <-
 
 #' @export
 prep.step_lag <- function(x, training, info = NULL, ...) {
+  if (!all(x$lag == as.integer(x$lag))) {
+    rlang::abort("step_lag requires 'lag' argument to be integer valued.")
```
simonpcouch, line 95:
```suggestion
    rlang::abort("step_lag() requires 'lag' argument to be integer-valued.")
```

simonpcouch: I'm on board for moving this from `bake()` to `prep()`. Could we document in this PR (here is fine) why that is?

simonpcouch (approved):
Nice! :)

R/lag.R
EmilHvitfeldt: There are 3 places where we can put `abort()` in a step. `step_lag()`, `prep.step_lag()` and `bake.step_lag()`.

If we put in in `step_lag()` you won't be able to run the following code

```r
library(recipes)
recipe(~., data = mtcars) |>
  step_lag(mpg, lag = 0.5)
```

If you put it in `prep.step_lag()`, we get the error once we call `prep()` or `fit()` if it were in a workflow.

And if it is put in `bake.step_lag()`, then we get the error every time we try to `bake()` or `predict()` with it. 

I don't like the first option, because it means that you can't create a wrong specification. And it makes testing harder.

I don't like the last option, because then code would run every time we try to apply it to new data. Which is redundant since the input isn't changed from time to time. Hence why I put it in `prep.step_lag()`

PR #1182, opened 2023-08-02
loopify step_dummy()

EmilHvitfeldt:

This PR does a small refactor to `step_dummy()` to bring them up to the same standard as https://github.com/tidymodels/recipes/pull/1156.

### DIFF
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -258,14 +258,8 @@ warn_new_levels <- function(dat, lvl, details = NULL) {
 
 #' @export
 bake.step_dummy <- function(object, new_data, ...) {
-  check_new_data(names(object$levels), object, new_data)
-
-  # If no terms were selected
-  if (length(object$levels) == 0) {
-    return(new_data)
-  }
-
   col_names <- names(object$levels)
+  check_new_data(col_names, object, new_data)
 
   ## `na.action` cannot be passed to `model.matrix` but we
   ## can change it globally for a bit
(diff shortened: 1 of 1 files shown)


R/dummy.R
```diff
@@ -258,85 +258,75 @@ warn_new_levels <- function(dat, lvl, details = NULL) {
 
 #' @export
 bake.step_dummy <- function(object, new_data, ...) {
-  check_new_data(names(object$levels), object, new_data)
-
-  # If no terms were selected
-  if (length(object$levels) == 0) {
```
simonpcouch, line 264: It looks like we didn't re-introduce the early return that was here?

R/dummy.R
```diff
@@ -258,85 +258,75 @@ warn_new_levels <- function(dat, lvl, details = NULL) {
...
         na.action = na.pass
       )
 
     indicators <-
       model.matrix(
-        object = object$levels[[i]],
+        object = levels,
         data = indicators
       )
-    indicators <- as_tibble(indicators)
```
simonpcouch, line 320: 😎

R/dummy.R
EmilHvitfeldt: Not needed since we are looping over the selected variables https://github.com/tidymodels/recipes/pull/1182/files#diff-db374e7fd2fd247f43cade403bf9b66b0abad4ff7dffc9025b2f06e45d660fd0R270

simonpcouch: Feels worth not touching the exit handlers unless needed though, no?

EmilHvitfeldt: true, will put it back in

PR #1183, opened 2023-08-02
loopify step_classdist()

EmilHvitfeldt:

This PR does a small refactor to `step_classdist()` to bring them up to the same standard as https://github.com/tidymodels/recipes/pull/1156.

### DIFF
--- a/R/classdist.R
+++ b/R/classdist.R
@@ -230,34 +230,40 @@ mah_pooled <- function(means, x, cov_mat) {
 
 #' @export
 bake.step_classdist <- function(object, new_data, ...) {
-  if (length(object$objects[[1]]$center) == 0) {
+  col_names <- names(object$objects[[1]][[1]])
+  check_new_data(col_names, object, new_data)
+
+  if (length(col_names) == 0) {
     return(new_data)
   }
 
   if (object$pool) {
-    x_cols <- names(object$objects[["center"]][[1]])
-    check_new_data(x_cols, object, new_data)
-    res <- lapply(
+    new_values <- lapply(
       object$objects$center,
       mah_pooled,
-      x = new_data[, x_cols],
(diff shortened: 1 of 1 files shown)


R/classdist.R
```diff
@@ -230,34 +230,40 @@ mah_pooled <- function(means, x, cov_mat) {
...
-  res <- vec_cbind(new_data, res)
-  res <- remove_original_cols(res, object, x_cols)
-  res
+  new_values <- tibble::new_tibble(new_values)
+
+  new_names <- paste0(object$prefix, colnames(new_values))
+  colnames(new_values) <- new_names
+
+  new_values <- check_name(new_values, new_data, object, new_names)
+  new_data <- vec_cbind(new_data, new_values)
```
simonpcouch, line 264:
```suggestion
  new_data <- vctrs::vec_cbind(new_data, new_values)
```

R/classdist.R
```diff
@@ -230,34 +230,40 @@ mah_pooled <- function(means, x, cov_mat) {
 
 #' @export
 bake.step_classdist <- function(object, new_data, ...) {
-  if (length(object$objects[[1]]$center) == 0) {
+  col_names <- names(object$objects[[1]][[1]])
```
simonpcouch, line 233: Transitioning to selection by position has me a little 🤨 but this looks safe.

simonpcouch (approved):
Groovy!

R/classdist.R
EmilHvitfeldt: Yes, this wasn't ideal, but it is because the information is in two different places depending on `pool`. But it just happens to be that `names(object$objects[[1]][[1]])` captures both places 😄

EmilHvitfeldt: also, the positioning is just "first"

Issue #1181, opened 2023-08-01
Switch away from using `map_dfc()` and `map_dfr()`


EmilHvitfeldt:
Deprecated `map_dfc()`. Let's open a new issue to rewrite this, as this is unrelated to this PR. :)

_Originally posted by @simonpcouch in https://github.com/tidymodels/recipes/pull/1156#discussion_r1281180620_

PR #1204, opened 2023-09-27
move away from purrr::map_dfc() and purrr::map_dfr()

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/1181


simonpcouch (approved):
Looking good!

PR #1209, opened 2023-09-28
document that roles functions are applied before steps and checks

### LINKED ISSUE #778: Update roles not working when variables are added in a previous recipe step
## The problem

When creating features during recipe steps, we can't update or change the role of new variables if they weren't available in the template dataset. 

### Reproducible example

```r
data(ad_data, package = "modeldata")


ad_rec <- recipe(Class ~ tau + VEGF, data = ad_data) %>%
  step_mutate(tau_2 = tau*2, tau_3 = tau*3) %>%
  add_role(tau_2, new_role="test_role")

```

Output

```
Error: Can't subset columns that don't exist.
x Column `tau_2` doesn't exist.
```

EmilHvitfeldt:
to close #778

This should really have been done earlier, we had a good amount of people confused 😄

### DIFF
--- a/R/roles.R
+++ b/R/roles.R
@@ -11,22 +11,24 @@
 #' `remove_role()` eliminates a single existing role in the recipe.
 #' @param recipe An existing [recipe()].
 #'
-#' @param ... One or more selector functions to choose which variables are
-#'   being assigned a role. See [selections()] for more details.
-#'
+#' @param ... One or more selector functions to choose which variables are being
+#'   assigned a role. See [selections()] for more details.
 #' @param new_role A character string for a single role.
-#'
 #' @param new_type A character string for specific type that the variable should
-#' be identified as. If left as `NULL`, the type is automatically identified
-#' as the _first_ type you see for that variable in `summary(recipe)`.
-#'
+#'   be identified as. If left as `NULL`, the type is automatically identified
+#'   as the _first_ type you see for that variable in `summary(recipe)`.
 #' @param old_role A character string for the specific role to update for the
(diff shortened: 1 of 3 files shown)


R/roles.R
```diff
@@ -11,23 +11,25 @@
...
-#' variables have only a single role.
+#'   variables selected by `...`. `update_role()` accepts a `NULL` as long as
+#'   the variables have only a single role.
 #'
 #' @return An updated recipe object.
 #'
 #' @details
 #'
+#' `update_role()`, `add_role()` and `remove_role()` will be applied on a recipe
+#' before any of the steps or checks. Regardless on where they are located in
```
simonpcouch, line 29:
```suggestion
#' before any of the steps or checks, regardless of where they are located in
```

R/roles.R
```diff
@@ -11,23 +11,25 @@
...
+#' pipe. This means that roles can only be changed on the original data supplied
```
simonpcouch, line 30:
```suggestion
#' position. This means that roles can only be changed on the original data supplied
```

simonpcouch: pipe -> position since using the pipe isn't required.

simonpcouch (commented):
Does:

> This means that roles can only be changed on the original data supplied to `recipe()`.

...refer to columns? Like, one can't update roles for columns that the recipe creates via these functions, but only via the `role` argument? If that's the case, I'd suggest something like:

> This means that roles can only be changed with these three functions for columns that are already present in the original data supplied to `recipe()`. See the `role` argument in some step functions to update roles for columns created by steps.

Issue #907, opened 2022-02-17
updated Selecting Variables vignette 


EmilHvitfeldt:
The vignette should be updated to include information about `all_predictor_*()` functions and the new things can was made possible with the change to `recipes_eval_select()` and being able to select 1 variables in most steps.

EmilHvitfeldt:
I'm going to assume past Emil meant `select 0 variables in most steps` since that is in fact interesting

PR #1212, opened 2023-09-29
updating selecting variants vignette

EmilHvitfeldt:
To close https://github.com/tidymodels/recipes/issues/907

This PR:

- adds talk about new fancy selection helper functions
- Adds note about empty selections
- switches to using penguins since they have integers and doubles

### DIFF
--- a/vignettes/Selecting_Variables.Rmd
+++ b/vignettes/Selecting_Variables.Rmd
@@ -86,3 +86,4 @@ with_dummy
 
 One important aspect about selecting variables in steps is that the variable names and types may change as steps are being executed. In the above example, `Records` is a factor variable before the step is executed. Afterwards, `Records` is gone and the binary variable `Records_yes` is in its place. One reason to have general selection routines like `all_predictors()` or `contains()` is to be able to select variables that have not be created yet. 
 
+All steps support empty selections. Meaning that if `all_date_predictors()` is used in a step, and no date variables was found the in the data set, then the step is applied without error. The calculations inside the step will be skipped. This allows for quite relaxed recipes as you don't have to make sure that the variables exists at that point in the recipe.
--- a/vignettes/Selecting_Variables.Rmd
+++ b/vignettes/Selecting_Variables.Rmd
@@ -33,7 +33,7 @@ The manual pages for `?selections` and  `?has_role` have details about the avail
  
 To illustrate this, the palmer penguins data will be used: 
 
-```{r credit}
+```{r penguins}
 library(recipes)
 library(modeldata)
 
@@ -52,7 +52,7 @@ summary(rec, original = TRUE)
 
 This shows the types and roles. Each variable can have one or more types, so we can printing them out seperately
 
-```{r var_info_orig}
+```{r var_info_orig_type}
 summary(rec, original = TRUE)$type
 ```
 
--- a/vignettes/Selecting_Variables.Rmd
+++ b/vignettes/Selecting_Variables.Rmd
@@ -2,7 +2,7 @@
 title: "Selecting variables"
 output: rmarkdown::html_vignette
 description: |
-  You can select which variables or features  should be used in recipes. This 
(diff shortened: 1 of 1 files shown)


vignettes/Selecting_Variables.Rmd
```diff
@@ -50,29 +50,49 @@ Before any steps are used the information on the original variables is:
...
+One important aspect about selecting variables in steps is that the variable names and types may change as steps are being executed. In the above example, `sex` is a factor variable, if `step_dummy()` was used on it, then `sex` would be done gone and the binary variable `sex_male` is in its place. One reason to have general selection routines like `all_predictors()` or `contains()` is to be able to select variables that have not be created yet. 
 
+All steps support empty selections. Meaning that if `all_date_predictors()` is used in a step, and no date variables was found the in the data set, then the step is applied without error. The calculations inside the step will be skipped. This allows for quite relaxed recipes as you don't have to make sure that the variables exists at that point in the recipe.
```
topepo, line 98:
```suggestion
All steps in the recipes package support empty selections. Meaning that if `all_date_predictors()` is used in a step, and no date variables was found the in the data set, then the step is applied without error. The calculations inside the step will be skipped. This allows for quite relaxed recipes as you don't have to make sure that the variables exists at that point in the recipe.
```

vignettes/Selecting_Variables.Rmd
```diff
@@ -50,29 +50,49 @@ Before any steps are used the information on the original variables is:
...
+dummied <- prep(dummied, training = penguins)
+with_dummy <- bake(dummied, new_data = penguins)
 with_dummy
 ```
 
-`Status` is unaffected. 
+`body_mass_g` is unaffected. 
 
-One important aspect about selecting variables in steps is that the variable names and types may change as steps are being executed. In the above example, `Records` is a factor variable before the step is executed. Afterwards, `Records` is gone and the binary variable `Records_yes` is in its place. One reason to have general selection routines like `all_predictors()` or `contains()` is to be able to select variables that have not be created yet. 
+One important aspect about selecting variables in steps is that the variable names and types may change as steps are being executed. In the above example, `sex` is a factor variable, if `step_dummy()` was used on it, then `sex` would be done gone and the binary variable `sex_male` is in its place. One reason to have general selection routines like `all_predictors()` or `contains()` is to be able to select variables that have not be created yet.
```
topepo, line 96:
```suggestion
One important aspect of selecting variables in steps is that the variable names and types may change as steps are being executed. In the above example, `sex` is a factor variable, if `step_dummy()` was used on it, then `sex` would be removed and the binary variable `sex_male` is in its place. One reason to have general selection routines like `all_predictors()` or `contains()` is to be able to select variables that have not been created yet. 
```

vignettes/Selecting_Variables.Rmd
```diff
@@ -50,29 +50,49 @@ Before any steps are used the information on the original variables is:
...
 ```
 
-We can add a step to compute dummy variables on the non-numeric data after we impute any missing data:
+This shows the types and roles. Each variable can have one or more types, so we can printing them out seperately
+
+```{r var_info_orig_type}
+summary(rec, original = TRUE)$type
+```
+
+Notice that integer variables have roles `"integer"` and `"numeric"`, and the factor variable have roles `"factor"`, `"unordered"`, `"nominal"`. This allows for some neat selections where the selector `all_numeric()` select double and integer variables, and more specific selectors such as `all_integer()` only selects integer variables. Full hierachy of types can be seen in `?has_role`.
```
topepo, line 59:
```suggestion
Notice that integer variables have roles `"integer"` and `"numeric"`, and the factor variables have roles `"factor"`, `"unordered"`, `"nominal"`. This allows for some neat selections where the selector `all_numeric()` select double and integer variables, and more specific selectors such as `all_integer()` only select integer variables. A full hierarchy of types can be seen in `?has_role`.
```

Issue #1237, opened 2023-10-11
Make better use of {cli}


EmilHvitfeldt:
This is a collection issue I'll use to keep track of all the {cli} based changes that is going to happen over the next couple of months.

## Infrastructure

- `check_new_data()` https://github.com/tidymodels/recipes/pull/1014 (old)
- `check_type()` https://github.com/tidymodels/recipes/pull/1235
- `recipes_eval_select()` https://github.com/tidymodels/recipes/pull/1238
- `check_name()` https://github.com/tidymodels/recipes/pull/1240
- `get_keep_original_cols()` https://github.com/tidymodels/recipes/pull/1241
- case weights functions https://github.com/tidymodels/recipes/pull/1242
- `check_nominal_type()` https://github.com/tidymodels/recipes/pull/1245
- smaller functions in `formula.R` and `misc.R` https://github.com/tidymodels/recipes/pull/1244
- roles functions in `roles.R` https://github.com/tidymodels/recipes/pull/1246
- everything in `recipe.R` https://github.com/tidymodels/recipes/pull/1251
- everything non-`abort()` https://github.com/tidymodels/recipes/pull/1264

## Steps
- https://github.com/tidymodels/recipes/pull/1255
- https://github.com/tidymodels/recipes/pull/1262

EmilHvitfeldt:
I have been using `.var` to denote variables in the data set, but I'm wondering if `.field` is better?

![Screenshot 2023-10-13 at 5 09 52 PM](https://github.com/tidymodels/recipes/assets/14034784/6038b89d-f308-4e3b-978f-d30f19292314)

PR #1242, opened 2023-10-12
update case weights functions to use cli

EmilHvitfeldt:
Ref: https://github.com/tidymodels/recipes/issues/1237

### DIFF
--- a/R/case_weights.R
+++ b/R/case_weights.R
@@ -113,8 +113,7 @@ too_many_case_weights <- function(x, call = rlang::caller_env()) {
     c(
       "!" = "There should only be a single column with the role \\
       {.code case_weights}.",
-      "i" = "{cli::qty(n)}In these data, there {?is/are} {n} column{?s}: \\
-      {.var {x}}."
+      "i" = "In these data, there are {n} columns: {.var {x}}."
     ),
     call = call
   )
--- a/R/case_weights.R
+++ b/R/case_weights.R
@@ -82,20 +82,21 @@ NULL
 get_case_weights <- function(info, .data) {
   wt_col <- info$variable[info$role == "case_weights" & !is.na(info$role)]
 
-
   if (length(wt_col) == 1) {
     res <- .data[[wt_col]]
     if (!is.numeric(res)) {
-      rlang::abort(
-        paste0(
-          "Column ", wt_col, " has a 'case_weights' role but is not numeric."
-        )
+      cli::cli_abort(
+        c(
+          "x" = "{.arg {wt_col}} has a {.code case_weights} role,\\
+                 but is not numeric.",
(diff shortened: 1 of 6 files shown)


R/case_weights.R
```diff
@@ -79,36 +80,42 @@ NULL
...
-        paste0(
-          "Column ", wt_col, " has a 'case_weights' role but is not numeric."
-        )
+      cli::cli_abort(
+        c(
+          "x" = "{.arg {wt_col}} has a {.code case_weights} role,\\
+                 but is not numeric.",
+          "i" = "{.arg {wt_col}} is {.obj_type_friendly {wt_col}}."
+          ),
+        call = call
```
EmilHvitfeldt, line 95: tried for like 15 minutes, and was not able to reach this abort.

It is not the worst thing to keep i think? since it is just guarded behind a `is.numeric()` which is fast.

simonpcouch: All good, thanks for the note. :)

R/case_weights.R
```diff
@@ -79,36 +80,42 @@ NULL
...
-      "In these data, there are ", n, " columns."
-    )
+too_many_case_weights <- function(x, call = rlang::caller_env()) {
+  n <- length(x)
+
+  cli::cli_abort(
+    c(
+      "!" = "There should only be a single column with the role \\
+      {.code case_weights}.",
+      "i" = "In these data, there are {n} column{?s}: {.var {x}}"
```
simonpcouch, line 116:
```suggestion
      "i" = "{cli::qty(n)}In these data, there {?is/are} {n} column{?s}: {.var {x}}"
```

So that we correctly pluralize is/are. Could we add a test case for multiple bad case weights columns?

``` r
msg <- "{cli::qty(n)}In these data, there {?is/are} {n} column{?s}: {.var {x}}"

x <- c("hey", "there")
n <- length(x)

cli::cli_abort(msg)
#> Error:
#> ! In these data, there are 2 columns: `hey` and `there`
#> Backtrace:
#>     ▆
#>  1. └─cli::cli_abort(msg)
#>  2.   └─rlang::abort(...)

x <- c("hey")
n <- length(x)

cli::cli_abort(msg)
#> Error:
#> ! In these data, there is 1 column: `hey`
#> Backtrace:
#>     ▆
#>  1. └─cli::cli_abort(msg)
#>  2.   └─rlang::abort(...)
```

<sup>Created on 2023-10-13 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I would maybe suggest adding a period at the end of this, too, but not sure if there's actually a grammar argument there.

simonpcouch (approved):
🕺

R/case_weights.R
EmilHvitfeldt: I just realized that this function is only called for `n>1` so it should always be called with pluralize

PR #1245, opened 2023-10-13
use cli in check_nominal_type()

EmilHvitfeldt:
ref: https://github.com/tidymodels/recipes/issues/1237

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -617,17 +617,13 @@ check_nominal_type <- function(x, lvl) {
     was_factor <- fac_ref_cols[!(fac_ref_cols %in% fac_act_cols)]
 
     if (length(was_factor) > 0) {
-      rlang::warn(
-        paste0(
-          " There ",
-          ifelse(length(was_factor) > 1, "were ", "was "),
-          length(was_factor),
-          ifelse(length(was_factor) > 1, " columns ", " column "),
-          "that ",
-          ifelse(length(was_factor) > 1, "were factors ", "was a factor "),
-          "when the recipe was prepped:\n ",
-          paste0("'", was_factor, "'", collapse = ", "),
-          ".\n This may cause errors when processing new data."
+      cli::cli_warn(
+        c(
+          "!" = " There {?w/was/were} {length(was_factor)} column{?s} that \\
+                {?w/was a factor/were factors} when the recipe was prepped: \\
+                ",
+          "*" = "{.and {.var {was_factor}}}",
+          "i" = "This may cause errors when processing new data."
         )
       )
     }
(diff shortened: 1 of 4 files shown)


R/misc.R
```diff
@@ -617,17 +617,13 @@ check_nominal_type <- function(x, lvl) {
...
-          length(was_factor),
-          ifelse(length(was_factor) > 1, " columns ", " column "),
-          "that ",
-          ifelse(length(was_factor) > 1, "were factors ", "was a factor "),
-          "when the recipe was prepped:\n ",
-          paste0("'", was_factor, "'", collapse = ", "),
-          ".\n This may cause errors when processing new data."
+      cli::cli_warn(
+        c(
+          "!" = " There {?w/was/were} {length(was_factor)} column{?s} that \\
```
simonpcouch, line 622:
```suggestion
          "!" = "There {?w/was/were} {length(was_factor)} column{?s} that \\
```

This raised an eyebrow, and I see no change in output after deleting it.

R/misc.R
```diff
@@ -617,17 +617,13 @@ check_nominal_type <- function(x, lvl) {
...
+                {?w/was a factor/were factors} when the recipe was prepped: \\
+                ",
+          "*" = "{.and {.var {was_factor}}}",
```
simonpcouch, line 625: Could we test this pluralization with a 1-length `was_factor`? Running locally, it looks like it works well, though. :)

R/misc.R
```diff
@@ -617,17 +617,13 @@ check_nominal_type <- function(x, lvl) {
...
-          ifelse(length(was_factor) > 1, " columns ", " column "),
-          "that ",
-          ifelse(length(was_factor) > 1, "were factors ", "was a factor "),
-          "when the recipe was prepped:\n ",
-          paste0("'", was_factor, "'", collapse = ", "),
-          ".\n This may cause errors when processing new data."
+      cli::cli_warn(
+        c(
+          "!" = " There {?w/was/were} {length(was_factor)} column{?s} that \\
+                {?w/was a factor/were factors} when the recipe was prepped: \\
```
simonpcouch, line 623:
```suggestion
                {?was a factor/were factors} when the recipe was prepped: \\
```

I believe we can drop this change both here and in the analogous part of the line above. The `?w` markup would only kick in when `length(was_factor) == 0`, which is never the case in here, but even in that case, just substituting in `"w"` is incorrect whereas the plural `"were factors"` reads fine.

simonpcouch (approved):
This one is so satisfying!

R/misc.R
EmilHvitfeldt: this was a typo :(

R/misc.R
EmilHvitfeldt: done in [`8bcdf16` (#1245)](https://github.com/tidymodels/recipes/pull/1245/commits/8bcdf16636b0f7c70626187f211a1d7c97693d13)

PR #1257, opened 2023-11-06
link to info on model formulas in error message

simonpcouch:
This PR follows up on https://github.com/tidymodels/parsnip/pull/1015, where I put together a help-page on model formulas that we could link to throughout tidymodels errors and docs. I think this check would be a great place to spot folks who need this information but haven't found it yet.

Made sure to think through the "cascade" when putting these PRs together. With merging this PR as-is, this would put as at:

* tune, then yardstick

and, independently, 

* parsnip, then recipes

### DIFF
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -279,9 +279,11 @@ inline_check <- function(x) {
     cli::cli_abort(c(
       x = "No in-line functions should be used here.",
       i = "{cli::qty(length(funs))}The following function{?s} {?was/were} \\
-          found:",
-      "*" = "{.and {.code {funs}}}",
-      i = "Use steps to do transformations instead."
+          found: {.and {.code {funs}}}.",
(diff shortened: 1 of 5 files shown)


R/recipe.R
```diff
@@ -279,9 +279,11 @@ inline_check <- function(x) {
     cli::cli_abort(c(
       x = "No in-line functions should be used here.",
       i = "{cli::qty(length(funs))}The following function{?s} {?was/were} \\
```
simonpcouch, line 281: In order to not increase the number of `i` entries further, I opted to collapse the itemization onto the line where it's introduced. I believe this happens in a few other places in recipes already!

simonpcouch:
dev deployment of those docs are [here](https://parsnip.tidymodels.org/dev/reference/model_formula.html)!

EmilHvitfeldt (approved):
Looking good!

PR #1263, opened 2023-11-17
fix: print_step not using width parameter

atusy:
Althou variety of `print.step_*()` passes its own `width` to `print_step()`, `print_step()` is not using it.

This PR fixes the problem.

Example of widths defined by `print.step_*()` are below

https://github.com/tidymodels/recipes/blob/17211e6dfc61516adf510cb27a543ce1bef44da9/R/dummy.R#L313-L314

https://github.com/tidymodels/recipes/blob/17211e6dfc61516adf510cb27a543ce1bef44da9/R/inverse.R#L110-L111

### DIFF
--- a/R/printing.R
+++ b/R/printing.R
@@ -46,7 +46,7 @@ print_step <- function(tr_obj = NULL,
     )
   )
 
-  width_diff <- cli::console_width() * 1 - width_title
+  width_diff <- width - width_title
 
   if (trained) {
     elements <- tr_obj


EmilHvitfeldt:
Hello @atusy! Thanks for looking into this, unfortunately this isn't an issue. The code is written the way it is because the old way of printing the steps with `printer()`

https://github.com/tidymodels/recipes/blob/6bc9cb04297d14a6aa712aceb0aa1c1dcef4b181/R/misc.R#L299-L302

used to use `width` as a way to figure out how wide to print. It wasn't working that well and it is now no longer needed as `print_step()` is able to calculate everything it needs itself.

The arguments are still there to avoid breaking stuff as we moved over the change

PR #1285, opened 2024-02-12
Fix for legacy recipes in `bake.step_log()` method

### LINKED ISSUE #1284: The `bake.step_log()` S3 method breaks legacy recipe objects
## The problem

The current (CRAN `v1.0.9`) versions of `bake.step_log()` breaks older (legacy) recipes made prior to (guessing) `v1.0.4`, >  1 year ago. When attempting to `bake()` a prepped recipe that has a `log_step`, the `bake()` S3 method _used to_ check for the appropriate columns using the `x$columns` entry of the object, however it _now_ looks for its names, i.e. `names(x$columns)`, which for older recipes without vector names will be `NULL`, and breaks backward compatibility of the object.

The line where this happens is: https://github.com/tidymodels/recipes/blob/main/R/log.R#L127

And it appears to have been introduced here: 6155183

Perhaps even worse, the `bake()` step does not fail when names are `NULL`, it skips the logging-loop and exits without modifying the any columns, and resulting in incorrect pre-processed data.

## Questions:

1. are there any future plans to address backward compatibility of older objects?
1. is there a recommended (preferred) workaround?
1. should breaking changes of legacy recipes be a consideration in semantic versioning?
1. should I modify the recipe object as necessary for it to function in `v1.0.9`? 

The last option is not ideal since in my use case, these recipes (~45) are production level objects that are frozen with their corresponding model objects and are tied to regulatory controlled customer products.

## Reprex

I do not have a `reprex` _per se_, but there is a simple workaround:

```r
update_legacy_logstep <- function(obj) {
  idx <- which(vapply(obj$steps, inherits, what = "step_log", NA))
  for (i in seq_len(idx)) {
    step <- obj$steps[[i]]
    if (is.null(names(step$columns))) {
      names(obj$steps[[i]]$columns) <- step$columns
    }
  }
  obj
}
```

stufield:
## Possible Bugfix

- Here is one option to allow legacy objects to still generate `bake()ed` objects in the new `recipes` syntax.
- Not sure if this fits in the larger plan for future upgrades and harmonizing with other steps in the recipe suite, however.
- Feel free to decline if this does not fit with the larger context.

## commit message:
- legacy recipes index the `columns` entry of `object` whereas newer recipes index `names(object$columns)`
- this breaks backward compatibility
- This bugfix prefers the 'new' syntax, but if `names(object)` is `NULL`, reverts to the 'old' syntax
- fixes #1284

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,8 @@
 
 ## Bug Fixes
 
+* Fixed bug where `step_log()` breaks legacy recipe objects by indexing `names(object)` in `bake()`. (@stufield, #1284)
+
 * Fixed bug where `step_factor2string()` if `strings_as_factors = TRUE` is set in `prep()`. (#317)
 
 * Fixed bug where `tidy.step_cut()` always returned zero row tibbles for trained recipes. (#1229)
--- a/R/log.R
+++ b/R/log.R
@@ -124,6 +124,7 @@ prep.step_log <- function(x, training, info = NULL, ...) {
 
 #' @export
 bake.step_log <- function(object, new_data, ...) {
+  # For backward compatibility #1284
   col_names <- names(object$columns) %||% object$columns
   check_new_data(col_names, object, new_data)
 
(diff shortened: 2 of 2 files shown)


EmilHvitfeldt:
Looking good!

Thanks for the PR, I'll start working on a patch release first thing tomorrow

PR #1293, opened 2024-03-27
Remove ellipsis dependency.

olivroy:
As ellipsis is deprecated (and imports rlang). https://rlang.r-lib.org/news/index.html#argument-intake-1-0-0

No change to commit from documenting.

Since recipes uses `@import rlang`, this will work as is.

### DIFF
--- a/R/selections.R
+++ b/R/selections.R
@@ -123,7 +123,7 @@ NULL
 #'
 #' This is a developer tool that is only useful for creating new recipes steps.
 #'
-#' @inheritParams ellipsis::dots_empty
+#' @inheritParams rlang::args_dots_empty
 #'
 #' @param quos A list of quosures describing the selection. This is generally
 #'   the `...` argument of your step function, captured with [rlang::enquos()]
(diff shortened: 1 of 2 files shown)


EmilHvitfeldt:
Thank you for finding this!

R/selections.R
```diff
@@ -172,7 +172,7 @@ NULL
 #' recipes_eval_select(quos, scat, info)
 recipes_eval_select <- function(quos, data, info, ..., allow_rename = FALSE,
                                 check_case_weights = TRUE, call = caller_env()) {
-  ellipsis::check_dots_empty()
+  check_dots_empty()
```
olivroy, line 175:
```suggestion
  check_dots_empty(call = call)
```

@EmilHvitfeldt actually, may be worth adding the call argument since `check_dots_empty()` accepts it and `recipes_eval_select()` already has it?

EmilHvitfeldt: normally I would agree, but in this case, `check_dots_empty()` should only be triggered if the developer uses `recipes_eval_select()` incorrectly, so I think it is fine that `check_dots_empty()` is the one highlighted

Issue #1282, opened 2024-02-01
declare required packages for step_mutate


topepo:
Add an option, perhaps called `.pkgs`, and update the `required_pkgs()` method for this step. See [this post](https://community.rstudio.com/t/adding-packages-to-the-recipes-required-pkgs/181216)/

PR #1319, opened 2024-05-28
add .pkgs argument to step_mutate()

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/1282

it is too much work for us to magically figure out what packages people use in `step_mutate()` so we ask them to do it

``` r
library(recipes)

rec <- recipe(~., data = mtcars) |>
  step_mutate(new = 2)

rec_pred <- rec |>
  prep()

required_pkgs(rec)
#> [1] "recipes"
required_pkgs(rec_pred)
#> [1] "recipes"


rec <- recipe(~., data = mtcars) |>
  step_mutate(new = 2, .pkgs = "stats")

rec_pred <- rec |>
  prep()

required_pkgs(rec)
#> [1] "recipes" "stats"
required_pkgs(rec_pred)
#> [1] "recipes" "stats"
```

<sup>Created on 2024-05-28 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>


R/mutate.R
```diff
@@ -84,19 +86,25 @@
 #'
 #' # The difference:
 #' tidy(qq_rec, number = 1)
-step_mutate <- function(recipe, ...,
+step_mutate <- function(recipe, 
+                        ...,
+                        .pkgs = character(),
```
EmilHvitfeldt, line 91: since it is after the dots, order does matter, and for me I think having the arguments the user is more likely to need early makes sense

DESCRIPTION
```diff
@@ -70,6 +70,6 @@ Config/Needs/website: tidyverse/tidytemplate
 Config/testthat/edition: 3
 Encoding: UTF-8
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.3.1
-Remotes:  
+RoxygenNote: 7.3.1.9000
```
simonpcouch, line 73: I generally have feelings on using dev roxygen since the version number doesn't map to a definite SHA. Any reason this is required here?

R/mutate.R
simonpcouch: Agreed.

R/mutate.R
```diff
@@ -6,6 +6,8 @@
 #' @inheritParams step_pca
 #' @inheritParams step_center
 #' @param ... Name-value pairs of expressions. See [dplyr::mutate()].
+#' @param .pkgs character vector, package names of functions used in
```
simonpcouch, line 9:
```suggestion
#' @param .pkgs Character vector, package names of functions used in 
```

R/mutate.R
```diff
@@ -84,19 +86,25 @@
...
+step_mutate <- function(recipe, 
+                        ...,
+                        .pkgs = character(),
                         role = "predictor",
                         trained = FALSE,
                         inputs = NULL,
                         skip = FALSE,
                         id = rand_id("mutate")) {
+  check_character(.pkgs)
+  recipes_pkg_check(required_pkgs.step_mutate(list(.pkgs = .pkgs)))
```
simonpcouch, line 98: Because of the way this interfaces with S3, I think this is a situation where snapshot testing with a non-existent `.pkgs` would be nice.

NEWS.md
```diff
@@ -14,6 +14,8 @@
 
 * Fixed documentation mistake where default value of `keep_original_cols` argument were wrong. (#1314)
 
+* `step_mutate()` gained `.pkgs` argument to specify what packages needs to be loaded for step to work. (#1282)
```
simonpcouch, line 17:
```suggestion
* `step_mutate()` gained `.pkgs` argument to specify what packages need to be loaded for step to work. (#1282)
```

Might be worth clarifying here why `step_mutate()` needs it but others don't.

simonpcouch (approved):
🚀

DESCRIPTION
EmilHvitfeldt: mistake on my part

R/mutate.R
EmilHvitfeldt: done in [test step_mutate() with .pkgs not](https://github.com/tidymodels/recipes/pull/1319/commits/43b1110e69ca1d4faea2282d4ec454c042610c9c)

Issue #935, opened 2022-03-17
Add check to prevent a variable to be outcome and predictor at the same time


EmilHvitfeldt:
There are no checks in `add_role()` right now, and the following is possible.

``` r
library(recipes)

recipe(mpg ~., data = mtcars) %>%
  add_role("mpg", new_role = "predictor")
#> Recipe
#> 
#> Inputs:
#> 
#>       role #variables
#>    outcome          1
#>  predictor         11
```

This feels like a pattern that shouldn't be possible and would almost always be a mistake.

PR #1321, opened 2024-05-28
make add_roles() error if result is both predictor and outcome

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/935

``` r
library(recipes)

recipe(mpg ~., data = mtcars) %>%
  add_role("mpg", new_role = "predictor")
#> Error in `add_role()`:
#> ! `mpg` cannot get "predictor" role as it is already has role "outcome".
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 * `recipe()` will now show better error when columns are misspelled in formula (#1283).
 
-* `add_role()` now errors if one columns has role `"outcome"` and `"predictor"`. (#935)
+* `add_role()` now errors if a column would simultaneously have roles `"outcome"` and `"predictor"`. (#935)
 
 * Significant speedup in `step_dummy()` when applied to many columns. (#1305)
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 * `recipe()` will now show better error when columns are misspelled in formula (#1283).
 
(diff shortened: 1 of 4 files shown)


R/roles.R
```diff
@@ -217,6 +217,24 @@ add_role <- function(recipe, ..., new_role = "predictor", new_type = NULL) {
...
+    }
+    
+    if (new_role == "outcome" && any(old_roles == "predictor")) {
+      cli::cli_abort(
+        "{.var {var}} cannot get {.val outcome} role as it is already \\
+        has role {.val predictor}."
+      )
+    }
+  }
+
```
EmilHvitfeldt, line 237: it felt cleaner to have 2 `cli::cli_abort()` calls than trying to do something clever to avoid it.

Also all the `"case_weights"` is handled elsewhere in the code, so no need to check for that here

NEWS.md
```diff
@@ -10,6 +10,8 @@
 
 * `recipe()` will now show better error when columns are misspelled in formula (#1283).
 
+* `add_role()` now errors if one columns has role `"outcome"` and `"predictor"`. (#935)
```
hfrick, line 13:
```suggestion
* `add_role()` now errors if a column would simultaneously have roles `"outcome"` and `"predictor"`. (#935)
```

R/roles.R
hfrick: Agree with erroring here 👍

hfrick (approved):
Definitely a qol improvement 🧁

Issue #1026, opened 2022-07-08
Add helper function to locate variables of specific types 


EmilHvitfeldt:
Many of the themis steps needs to know which variables are predictors, steps like `step_lencode_glm()` needs to know what the outcome is.

There should be a function here in {recipes} that provides this information. Not super complicated, but it should work correctly and be able to handle the NA roles as well.

(After Conf)

PR #1323, opened 2024-05-29
add recipes_name_predictors() and recipes_outcomes()

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/1026

I'm not attached to the name so feel free to suggest other names. It is not a user-facing function, so the name doesn't matter that much.

I have tried lately to use the prefix `recipes_` for these developer function names


hfrick (approved):
I think the name is fine! 👍

Issue #887, opened 2022-01-27
Function that removes step(s) from an existing recipe


walrossker:
## Feature

When testing multiple model types (e.g., in an ensemble workflow), one could create a baseline recipe that works for most of the models. Then if one model has slightly different pre-processing requirements (e.g., maybe you'd prefer to remove `step_dummy` in a `rand_forest` model or remove `step_rm` to include more predictors in an elastic net compared to a knn), a single call to this function could remove those steps and avoid the need to copy/paste the original recipe definition and delete certain steps. This helps to keep recipe definitions DRY.

## Reproducible Example

FWIW, I chose the verb `ignore_` to avoid confusion with 'remove' from `step_rm`. 

Happy to submit a pull request if this seems valuable.

```r
suppressPackageStartupMessages(devtools::load_all("."))
#> Loading recipes

rec <- recipe(Species ~ ., data = iris) %>%
  step_rm(Petal.Width, id = "rm_UDLut") %>%
  step_rm(starts_with("Sepal"), id = "custom_id")

rec %>% ignore_step("custom_id") # remove by custom id value
#> Recipe
#> 
#> Inputs:
#> 
#>       role #variables
#>    outcome          1
#>  predictor          4
#> 
#> Operations:
#> 
#> Variables removed Petal.Width
rec %>% ignore_step(1) # remove the first step
#> Recipe
#> 
#> Inputs:
#> 
#>       role #variables
#>    outcome          1
#>  predictor          4
#> 
#> Operations:
#> 
#> Variables removed starts_with("Sepal")
rec %>% ignore_step("rm") # remove all `step_rm` steps  (i.e., all steps here)
#> Recipe
#> 
#> Inputs:
#> 
#>       role #variables
#>    outcome          1
#>  predictor          4
#> 
#> Operations:
```

topepo:
We are looking it this a bit. 

If the recipe has been estimated (or partially estimated), this wouldn't be feasible since the recipe would have some critical data that can't be rolled back. We are verifying this. 

If the recipe has not been defined, this is pretty easy to do (just by subsetting the `recipes$steps` list). In this case, we could offer an easy api that takes `number` and/or `id` vectors as inputs and checks the recipe for having been trained.

walrossker:
Yeah the function I have now only works on unprepped recipes as you describe. I can imagine that removing steps from estimated recipes would be much more complicated.

PR #1324, opened 2024-05-29
add ignore_step function

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/887

Adds `ignore_step()`, with a similar interface as `tidy()` with respect to how we select, either using `number` or `id`.

I'm not in love with the name, so suggestions are welcome.


R/ignore_step.R
```diff
@@ -0,0 +1,71 @@
...
+      )
+    }
+    number <- which(id == step_ids)
+  }
+  
+  x$steps <- x$steps[-number]
+
+  if (length(x$steps) == 0) {
+    x["steps"] <- list(NULL)
+  }
```
EmilHvitfeldt, line 68: We do this to make sure that `ignore_step()` correctly returns a recipe with no steps

R/ignore_step.R
```diff
@@ -0,0 +1,71 @@
...
+    )
+  }
+  
+  n_steps <- length(x$steps)
+
+  if (n_steps == 0) {
+    cli::cli_abort("{.arg x} doesn't contain any steps to remove.")
+  }
+
+  arg <- rlang::check_exclusive(number, id)
```
hfrick, line 44: We could allow `ignore_step(<recipes without steps>)` and just have it do nothing. If we don't want that, I would swap these two checks. To me, it's more obvious then that we don't allow that because we always require you to provide what to remove (before we tell you that there is nothing to remove).

R/ignore_step.R
```diff
@@ -0,0 +1,71 @@
...
+#'   step_normalize(all_numeric_predictors()) %>%
+#'   step_pca(all_numeric_predictors(), id = "PCA")
+#' 
+#' ignore_step(rec, number = 1)
+#' 
+#' ignore_step(rec, number = 1:2)
+#' 
+#' ignore_step(rec, id = "PCA")
+#' @export
+ignore_step <- function(x, number, id) {
```
hfrick, line 31: I would give `number` and `id` some default values since only one of the two is required. `tidy()` uses `NA`.

tests/testthat/test-ignore_step.R
```diff
@@ -0,0 +1,108 @@
...
+  rec123 <- recipe(mpg ~ ., data = mtcars) %>%
+    step_dummy(all_nominal_predictors(), id = "dummy") %>%
+    step_impute_mean(all_numeric_predictors(), id = "impute_mean") %>%
+    step_normalize(all_numeric_predictors(), id = "normalize")
+
+  expect_equal(
+    ignore_attr = TRUE,
+    ignore_step(rec1234, number = 1),
+    rec234
+  )
```
hfrick, line 28: I'm surprised by the argument order here, why do you put `ignore_attr` up front?

hfrick (commented):
I would not use "ignore" in the name here because that implies the step(s) in question are still part of the recipe, they just don't get used -- but this does remove them. I think `remove_step()` would be the cleanest but if we want to avoid "remove" because of `step_rm()`, we could go with "undo" or "delete".

R/ignore_step.R
```diff
@@ -0,0 +1,71 @@
+#' Remove steps from recipe
+#'
+#' `ignore_step` will return a recipe without steps specified by the `number` or
+#' `id` argument.
+#' 
+#' @param x A `recipe` object.
```
hfrick, line 6:
```suggestion
#' @param x An untrained `recipe` object.
```

R/ignore_step.R
```diff
@@ -0,0 +1,71 @@
...
+#' @param number An integer vector, Denoting the positions of the steps that 
+#'   should be removed.
+#' @param id A character string. Denoting the `id` of the steps that should be
+#'   removed.
```
hfrick, line 10:
```suggestion
#' @param number An integer vector denoting the positions of the steps that 
#'   should be removed.
#' @param id A character string denoting the `id` of the steps that should be
#'   removed.
```

EmilHvitfeldt:
We are sitting on this for another week. Brief summary of meeting discussion:

- There is some doubt that this could lead to problems down the line AND/OR subtle bugs 
- feels like syntactic sugar
- Isn't a bad idea, but maybe better suited for an extension package
- can be imagined as part of a broader group of functions that modify recipes
    - swap steps, move steps up and down

sen-kaustav:
Hello,

I stumbled onto this feature when trying to figure out a way to modify my tidymodels recipe in my Shiny app. 

For context, I am building a Shiny app to allow end users to interactively fit and analyse GLM models.  As part of this, I am storing my `recipe()` as a `reactiveVal()` allowing it to be modified as the user interacts with the UI.  In this setup, being able to programmatically "reset" some parts of the recipe is really beneficial from a user experience perspective.  This allows them to go back and forth with trying out different pre-processing steps like how you would do in a coding / scripting environment.

As a workaround, I have currently copied over the ignore_step.R file to my project to be able to use this functionality.   I have also modified it slightly to allow the passing a recipe step name as a character string.  For example, the below will delete the `step_normalize()` step from the recipe:

```
rec <- recipe(mpg ~ ., data = mtcars) %>%
  step_dummy(all_nominal_predictors()) %>%
  step_impute_mean(all_numeric_predictors()) %>%
  step_normalize(all_numeric_predictors()) %>%
  step_pca(all_numeric_predictors(), id = "PCA")

ignore_step(rec, step_type = "step_normalize")
```
For reference, here's a link to the modified ignore_step.R file that I am using: [ignore_step.R](https://github.com/sen-kaustav/glm-fit/blob/64bdcad12063c2a701d200b1a96e4787b5c0f4b8/ignore_step.R)

However, reading through the previous comments I can see that there are still a few open design considerations.  Given that here hasn't been any further traction on this ticket, I am guessing there are no plans to include this in a future release?  The current workaround does the job for me but having it as part of the core recipes package (or an extension package) would definitely be ideal!

Thanks again for the original code!  It really helped me solve my problem 😄.

_P.S. I am also not a huge fan of the `ignore_step()` name.  `omit_step()` or `delete_step()` (as previously suggested) has a much better ring to it._

Issue #793, opened 2021-09-14
`check_training_set()` should probably loosely check `training` column types


DavisVaughan:
If you specify a `recipe()` with a particular data frame, then `prep()` with a different data frame, `check_training_data()` does a few checks to make sure the two are compatible. In particular, right now it checks that the column names of the template data frame used in `recipe()` also exist in the data you `prep()` with. However, it doesn't check the types at all, so you can end up with odd errors like this one:

``` r
library(recipes)

df <- tibble(x = 1, y = "1")

# Original template `df` has a character `y`
rec <- recipe(~ ., df) %>%
  step_bin2factor(where(is.numeric))

# But when we prep we have a numeric `y`
df <- tibble(x = 1, y = 1)

# Awkward error
prep(rec, training = df, fresh = TRUE)
#> Error: The variables should be numeric

# The tidyselect bits selected `x` and `y` because they are both numeric,
# but in the original template data, `y` was nominal, which is why we get this error
rec$var_info
#> # A tibble: 2 × 4
#>   variable type    role      source  
#>   <chr>    <chr>   <chr>     <chr>   
#> 1 x        numeric predictor original
#> 2 y        nominal predictor original
```

I wonder if `check_training_set()` should also check the types a little somehow. It could check that the columns supplied through `prep()` are numeric or nominal if the original template supplied in `recipe()` had that type.

DavisVaughan:
This isn't a huge deal in practice, because workflows/hardhat typically handle the type consistency ahead of time

EmilHvitfeldt:
slightly bigger deal because {workflows}/{hardhat} doesn't deal with this.

Basically, we are in trouble if the data that is pased to `data` argument in `recipe()` doesn't match the data set that is used to `prep`/`fit()`

``` r
library(recipes)
#> Loading required package: dplyr
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
#> 
#> Attaching package: 'recipes'
#> The following object is masked from 'package:stats':
#> 
#>     step

traning <- tibble(outcome = rnorm(1000), x = rnorm(1000), y = sample(letters, 1000, T))

rec1 <- recipe(outcome ~ ., traning) %>%
  step_bin2factor(where(is.numeric), - all_outcomes())

# Original template `traning` has a character `y`
[...]

PR #1330, opened 2024-06-05
check for ptype information in `prep()`

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/793

This PR has 2 parts

1. adding `recipes_ptype_validate()`
    - since the ordering of the columns in the ptype can't be guaranteed, we can only check names, classes and attributes. This is a little tricky so i felt it would be worth it to export in case other packages (such as workflows and vetiver) wanted to check for themselves
2. using `recipes_ptype_validate()` in `check_training_set()`

``` r
library(recipes)

df <- tibble(x = 1, y = "1")

# Original template `df` has a character `y`
rec <- recipe(~ ., df) %>%
  step_bin2factor(where(is.numeric))

# But when we prep we have a numeric `y`
df <- tibble(x = 1, y = 1)

prep(rec, training = df, fresh = TRUE)
#> Error in `prep()`:
#> ✖ The following variable has the wrong class:
#> • `y` must have class <numeric> not <character>.
```


NEWS.md
```diff
@@ -4,6 +4,8 @@
 
 * Developer helper function `recipes_ptype()` has been added, returning expected input data for `prep()` and `bake()` for a given recipe object. (#1329)
 
+* Developer helper function `recipes_ptype_validate()` has been added, to validate new data is compatible with recipe ptype.(#793)
```
simonpcouch, line 7:
```suggestion
* Developer helper function `recipes_ptype_validate()` has been added, to validate new data is compatible with recipe ptype. (#793)
```

R/developer.R
```diff
@@ -144,6 +144,11 @@
 #' `stage` argument. Can be used by functions that interact with recipes to 
 #' verify data is correct before passing it to `prep()` and `bake()`.
 #' 
+#' [recipes_ptype_validate()] checks that a recipe and its data are compatible
+#' using information extracted using [recipes_ptype()]. Can be used by functions 
+#' that interact with recipes to  verify data is correct before passing it to
```
simonpcouch, line 149:
```suggestion
#' that interact with recipes to verify data is correct before passing it to 
```

R/ptype.R
```diff
@@ -46,7 +46,7 @@
 #' @return A zero row tibble.
 #' @keywords internal
 #'
-#' @seealso [developer_functions]
+#' @seealso [developer_functions] [recipes_ptype_validate]
```
simonpcouch, line 49: These link to the same place, no?

R/misc.R
```diff
@@ -663,19 +663,12 @@ check_training_set <- function(x, rec, fresh, call = rlang::caller_env()) {
...
-    }
     if (!is_tibble(x)) {
-      x <- as_tibble(x[, vars, drop = FALSE])
-    } else {
-      x <- x[, vars]
+      x <- as_tibble(x)
     }
+    recipes_ptype_validate(rec, new_data = x, stage = "prep", call = call)
+
+    x <- x[, vars]
```
simonpcouch, line 671: This seems to work just fine, but I would have expected this'd be implemented by just replacing the previous 666-673 with the call to `recipes_ptype_available()` and leaving the conditional subset as-is. Any reason why we did it this way?

R/ptype.R
```diff
@@ -111,3 +111,90 @@ recipes_ptype <- function(x, ..., stage = "prep") {
...
+
+  if (!identical(lapply(old_ptype, class), lapply(new_ptype, class))) {
+    old_classes <- lapply(old_ptype, class)
+    new_classes <- lapply(new_ptype, class)
+
+    offenders <- purrr::map2_lgl(old_classes, new_classes, identical)
+    offenders <- col_names[!offenders]
+
+    msg <- c("x" = "{cli::qty(offenders)} The following variable{?s} \\
+                   has the wrong class:")
```
simonpcouch, line 169:
```suggestion
    msg <- c("x" = "{cli::qty(offenders)} The following variable{?s} \\
                   {?has/have} the wrong class:")
```

Not sure if this order is right, but does need pluralization.

R/ptype.R
```diff
@@ -111,3 +111,90 @@ recipes_ptype <- function(x, ..., stage = "prep") {
...
+
+    col_msg <- paste0(
+      "{.var {offenders[",
+      seq_along(offenders),
+      "]}} must have class {.cls {new_classes[offenders[",
+      seq_along(offenders),
+      "]]}} not {.cls {old_classes[offenders[",
```
simonpcouch, line 176:
```suggestion
      "]]}}, not {.cls {old_classes[offenders[",
```

R/ptype.R
```diff
@@ -111,3 +111,90 @@ recipes_ptype <- function(x, ..., stage = "prep") {
...
+    old_classes <- lapply(old_ptype, class)
+    new_classes <- lapply(new_ptype, class)
+
+    offenders <- purrr::map2_lgl(old_classes, new_classes, identical)
+    offenders <- col_names[!offenders]
+
+    msg <- c("x" = "{cli::qty(offenders)} The following variable{?s} \\
+                   has the wrong class:")
+
+    col_msg <- paste0(
```
simonpcouch, line 171:
```suggestion
    # Use `paste0()` rather than typical glue syntax to intentionally duplicate messages
    col_msg <- paste0(
```

Tripped up for a second on first read of why this was written in the way that it was.

R/ptype.R
```diff
@@ -111,3 +111,90 @@ recipes_ptype <- function(x, ..., stage = "prep") {
...
+    
+  if (!identical(lapply(old_ptype, attributes), lapply(new_ptype, attributes))) {
+    old_attributes <- lapply(old_ptype, attributes)
+    new_attributes <- lapply(new_ptype, attributes)
+  
+    offenders <- purrr::map2_lgl(old_attributes, new_attributes, identical)
+    offenders <- col_names[!offenders]
+  
+    msg <- c("x" = "{cli::qty(offenders)} The following variable{?s} \\
+                   has the wrong attributed: {.var {offenders}}.")
```
simonpcouch, line 194:
```suggestion
                   {?has/have} the wrong attributes: {.var {offenders}}.")
```

Again, not sure if I have the right pluralization order here.

simonpcouch: In what cases do these attributes matter? Do we need to enforce that they not change?

Especially given that the message doesn't diagnose which attributes differ, I think I'd be pretty lost and frustrated if I saw this message.

tests/testthat/test-other.R
simonpcouch, line 1: The changes to this file might be a helpful place to leave a self-review comment explaining the reasoning.

tests/testthat/_snaps/ptype.md
```diff
@@ -7,3 +7,55 @@
...
+---
+
+    Code
+      recipes_ptype_validate(rec, data_new)
+    Condition
+      Error:
+      x The following variables has the wrong class:
+      * `id` must have class <numeric> not <integer>.
+      * `x2` must have class <integer> not <factor>.
+      * `y` must have class <numeric> not <integer>.
```
simonpcouch, line 44: For this type check specifically, I do wonder if we may be unintentionally putting users in a place where they didn't actually choose to do any coercion but some function of ours did it under the hood and they can no longer prep/bake. 

Let's keep an eye on extratests once this is merged.

simonpcouch (commented):
This looks great! Leaving review as "Comment" so I can take one last look at snaps after comments have been addressed.

R/ptype.R
EmilHvitfeldt: Examples would be time zone in datetime variables and factor levels are both stored as attributes. Which i think should both be checked

``` r
as.POSIXct(0, origin = "1960-01-01") |> attributes()
#> $tzone
#> [1] ""
#> 
#> $class
#> [1] "POSIXct" "POSIXt"
as.POSIXct(0, origin = "1960-01-01", tz = "GMT") |> attributes()
#> $tzone
#> [1] "GMT"
#> 
#> $class
#> [1] "POSIXct" "POSIXt"

factor(1:5) |> attributes()
#> $levels
#> [1] "1" "2" "3" "4" "5"
#> 
#> $class
#> [1] "factor"
factor(1:4) |> attributes()
#> $levels
#> [1] "1" "2" "3" "4"
#> 
#> $class
#> [1] "factor"
```

EmilHvitfeldt: i do agree that the error isn't great, the problem i ran into was it was hard to explain the difference as attributes could be anything

R/misc.R
EmilHvitfeldt: because `recipes_ptype_validate()` expects `x` to be a data.frame. So it didn't work when matrices were passed to `x`.

I have plans to streamline this in https://github.com/tidymodels/recipes/issues/1333

R/ptype.R
EmilHvitfeldt: `recipes_ptype_validate()` links to `recipes_ptype()` and `recipes_ptype()` links to `recipes_ptype_validate()` 😄

tests/testthat/_snaps/ptype.md
EmilHvitfeldt: 🤞 

I really hope we don't, because it would be good if we did 😅

tests/testthat/test-other.R
EmilHvitfeldt: the change here is a result of this test uses a mix of character and factor input in a way we disallow now. so the "fixed" test feels off

R/ptype.R
EmilHvitfeldt: what about?
- special case errors for factors and datetimes (most common)
- then more generic error, with code snippet as to how to figure out what went wrong?

simonpcouch: Sure, that sounds good!

simonpcouch (approved):
Looking good!🚀 Just that one comment on attribute checking.

PR #1359, opened 2024-08-15
Improve error messaging for `step_cut()` with missing data

### LINKED ISSUE #1351: uninformative error when `step_cut()` is applied to column with missing values
The error comes from `create_full_breaks()` and should properly be handled in here as well.

``` r
library(recipes)

recipe(~ ., data = mtcars) %>%
  step_cut(mpg, breaks = 20) %>%
  prep()
#> 
#> ── Recipe ──────────────────────────────────────────────────────────────────────
#> 
#> ── Inputs
#> Number of variables by role
#> predictor: 11
#> 
#> ── Training information
#> Training data contained 32 data points and no incomplete rows.
#> 
#> ── Operations
#> • Cut numeric for: mpg | Trained

mtcars[1, 1] <- NA

recipe(~ ., data = mtcars) %>%
  step_cut(mpg, breaks = 20) %>%
  prep()
#> Error in `step_cut()`:
#> Caused by error in `if (min(var) < min(breaks)) ...`:
#> ! missing value where TRUE/FALSE needed
```

<sup>Created on 2024-07-19 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

jrwinget:
Fixes #1351

Updated the `create_full_breaks()` function to provide a more informative error message when `step_cut()` is applied to a column containing missing values (`NA`). The function now checks for missing data and warns users appropriately.


EmilHvitfeldt:
Looking great! would you kindly add some snapshot tests using `extract_snapshot()` in https://github.com/tidymodels/recipes/blob/main/tests/testthat/test-cut.R.

Those tests should be placed above https://github.com/tidymodels/recipes/blob/main/tests/testthat/test-cut.R#L171

jrwinget:
Sure thing - thanks for the feedback!

jrwinget:
I've added the snapshot tests and pushed the changes. Please take a look when you have a moment. Thanks again!!

topepo:
Thanks for the contribution!

PR #1360, opened 2024-08-15
Add mday to `features` argument in `step_date()`

### LINKED ISSUE #1211: Add mday to step_date
The step_date function currently does not support the day of month field.  I would propose adding feature = 'mday' to step_date which would map to ```lubridate::mday```.  I would expect this code to add date_mday with a value of 28.  

I think the timezone could be handled by the locale argument.

```r
data <- data.frame(date = as.POSIXct('2023-09-28'))
recipes::recipe(data) |> 
  recipes::step_date(date, features = 'mday') |> 
  recipes::prep() |> 
  recipes::bake(data)
```

Edgar-Zamora:
Closes #1211


EmilHvitfeldt:
Very well done! looks perfect!

PR #1361, opened 2024-08-15
fix `step_novel` examples

### LINKED ISSUE #1248: step_novel Example looks out of date
## The problem

I'm trying to reproduce the example in the `step_novel` [function documentation](https://recipes.tidymodels.org/reference/step_novel.html#ref-examples). However when assigning a new value to the `city` predictor as per the example, instead of a new factor being added to the data set, it's automatically converted to `NA`. 

The example loses relevance after this since `step_novel` doesn't come into play. I suggest that the example is reviewed.

(Incidentally I'm having my own troubles with `step_novel` but I haven't built yet a reprex example I can show, I'll follow up with that).

Edgar-Zamora:
closes #1248

### DIFF
--- a/R/novel.R
+++ b/R/novel.R
@@ -49,8 +49,13 @@
 #'
 #' sacr_tr <- Sacramento[1:800, ]
 #' sacr_te <- Sacramento[801:806, ]
+#'
+#' # Without converting the predictor to a character, the new level would be converted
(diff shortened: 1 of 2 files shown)


EmilHvitfeldt:
Thank you! it looks great!

Issue #1343, opened 2024-06-13
Roadmap: recipes support for sparse tibbles


EmilHvitfeldt:
This issue is purely based on the integration part of using sparse tibbles in recipes as input/output. A different roadmap issue will be posted about having individual steps produce sparse columns.

What we need:

- [x] `recipe()` to take sparse tibbles as data https://github.com/tidymodels/recipes/pull/1364
- [x] `recipe()` to take {Matrix} sparse matrix as data https://github.com/tidymodels/recipes/pull/1367
  - turn them into sparse tibbles early
- [x] `prep()` to take sparse tibbles as data https://github.com/tidymodels/recipes/pull/1366
- [x] `prep()` to take {Matrix} sparse matrix as data https://github.com/tidymodels/recipes/pull/1368
  - turn them into sparse tibbles early
- [x] `bake()` to take sparse tibbles as data https://github.com/tidymodels/recipes/pull/1366
- [x] `bake()` to take {Matrix} sparse matrix as data https://github.com/tidymodels/recipes/pull/1369
  - turn them into sparse tibbles early
- [x] Document how sparse tibbles work within a recipe https://github.com/tidymodels/recipes/pull/1370
  - not all transformations are sparse safe

Adding all of this will give us
- standalone usage of sparse matrices in {recipes} 
- everything it needs to be able to work with the rest of {tidymodels} in regards to sparse tibbles

PR #1367, opened 2024-09-09
make `recipe()` work with sparse matrices

EmilHvitfeldt:
Ref: https://github.com/tidymodels/recipes/issues/1343


R/recipe.R
```diff
@@ -12,6 +12,13 @@ recipe <- function(x, ...) {
 #' @rdname recipe
 #' @export
 recipe.default <- function(x, ...) {
+
+  # Doing this here since it should work for all types of Matrix classes
```
EmilHvitfeldt, line 16: since it is a S4 class, I found this way of handling it the cleanest. It happens when you go `recipe(data, ...)`.

simonpcouch (approved):
Looking good!🦊 Just need to address:

```
❯ checking dependencies in R code ... WARNING
  '::' or ':::' import not declared from: ‘methods’
```

PR #1369, opened 2024-09-10
make `bake()` work with sparse matrices

EmilHvitfeldt:
Ref: https://github.com/tidymodels/recipes/issues/1343


simonpcouch (approved):
💯💅💪

PR #1370, opened 2024-09-10
document use of sparse data in recipes

EmilHvitfeldt:
ref: https://github.com/tidymodels/recipes/issues/1343

This documents the use of sparse data in the interface functions `recipe()`, `prep()`, and `bake()`. Documentation for steps will be added as those steps are modified to work with sparse data.


R/sparsevctrs.R
```diff
@@ -1,3 +1,22 @@
...
+#' 
+#' [recipe()], [prep()], and [bake()] all accept sparse tibbles from the 
+#' `sparsevctrs` package and sparse matrices from the `Matrix` package. Sparse
+#' matrices are converted to sparse tibbles internally as each step expects a 
+#' tibble as its input, and is expected to return a tibble as well.
+#' 
+#' Several steps work with sparse data. A step can either work with  sparse 
+#' data, ruin sparsity, or create sparsity. The documentation for each  step 
+#' will indicate whether it will work with sparse data or create sparse columns. 
+#' If nothing is listed it is assumed to ruin sparsity.
```
simonpcouch, line 11: Been sitting with these lines for a moment and not sure what I'd concretely suggest, but "ruin" doesn't feel like the word we want here (or its usage just needs clarification. Maybe we could link to some sparsevctrs docs here to clarify materializing sparse data?

R/sparsevctrs.R
```diff
@@ -1,3 +1,22 @@
...
+#' 
+#' Spare tibbles or data.frames will be return from [bake()] if sparse columns 
+#' are present in data, either from being generated in steps or because sparse 
+#' data was passed into [recipe()], [prep()], or [bake()].
```
simonpcouch, line 15:
```suggestion
#' Sparse tibbles or `data.frame`s will be returned from [bake()] if sparse columns 
#' are present in data, either from being generated in steps or because sparse 
#' data was passed into [recipe()], [prep()], or [bake()].
```

I think this is saying that sometimes, when supplied sparse data at bake, steps will materialize and others they won't? As in, `data.frame`s don't support sparsity? As written, it's unclear whether "Sparse" modifies "tibbles" or "tibbles or data.frames." Maybe 

> Supplied a sparse tibble, [bake()] will either return a `data.frame` or sparse tibble...

This ^^^ also assumes that, in "if sparse columns are present in data," we're not referring to sparse matrices.

simonpcouch (approved):
Wonderful!

I struggled a bit to figure out what was meant by the paragraphs in `?sparse_data`. I wrote a bit about what confused me but wasn't able to make very concrete suggestions as I'm not sure what the actual behavior of steps are in those cases. Leaving review as "Approve" just so as not to block you if you feel you're able to clarify those but feel free to re-request review!

R/sparsevctrs.R
EmilHvitfeldt: this section might be better once In have implemented it in a few steps and is able to make references

Issue #1363, opened 2024-09-09
Move away from `expect_error()`


EmilHvitfeldt:
same idea as https://github.com/tidymodels/parsnip/issues/1096

PR #1375, opened 2024-09-18
Use `expect_snapshot()` more

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/1363

This PR:

- switched from `expect_warning()` and `expect_message()` to `expect_snapshot()`
- removed classed `expect_error()` and switched to `expect_snapshot(error = TRUE)`
- Switched from `expect_error(..., NA)` to `expect_no_error()`. Can't do `expect_no_condition()` due to dplyr_regroup signal
- Switched rest of `expect_error()` to `expect_snapshot(error = TRUE)`

No bad snapshots were found in the switch 🙌

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -859,9 +859,8 @@ check_new_data <- function(req, object, new_data) {
   step_cls <- class(object)[1]
   step_id <- object$id
   cli::cli_abort(
-    "The following required {cli::qty(col_diff)} column{?s} {?is/are} \
-    missing from `new_data` in step '{step_id}': {col_diff}.",
-    class = "new_data_missing_column",
+    "The following required {cli::qty(col_diff)} column{?s} {?is/are} missing 
+    from `new_data`: {col_diff}.",
     call = rlang::call2(step_cls)
   )
 }
--- a/R/misc.R
+++ b/R/misc.R
@@ -860,7 +860,7 @@ check_new_data <- function(req, object, new_data) {
   step_id <- object$id
   cli::cli_abort(
     "The following required {cli::qty(col_diff)} column{?s} {?is/are} missing 
-    from `new_data`: {col_diff}.",
+    from {.arg new_data}: {col_diff}.",
     call = rlang::call2(step_cls)
   )
 }
--- a/R/mutate_at.R
+++ b/R/mutate_at.R
@@ -51,6 +51,10 @@ step_mutate_at <- function(recipe, ...,
                            inputs = NULL,
                            skip = FALSE,
                            id = rand_id("mutate_at")) {
+  if (rlang::is_missing(fn)) {
+    cli::cli_abort("Argument {.arg fn} must be specified.")
+  }
+
   add_step(
     recipe,
     step_mutate_at_new(
--- a/tests/testthat/test-colcheck.R
+++ b/tests/testthat/test-colcheck.R
@@ -5,32 +5,37 @@ rp1 <- recipe(mtcars, cyl ~ .)
 rp2 <- recipe(mtcars, cyl ~ mpg + drat)
 
 test_that("check_col works in the prep stage", {
-  expect_error(rp1 %>% check_cols(all_predictors()) %>% prep(), NA)
-  expect_error(rp2 %>% check_cols(all_predictors()) %>% prep(), NA)
-  expect_error(rp2 %>% check_cols(cyl, mpg, drat) %>% prep(), NA)
-  expect_error(rp2 %>% check_cols(cyl, mpg) %>% prep(), NA)
+  expect_no_error(rp1 %>% check_cols(all_predictors()) %>% prep())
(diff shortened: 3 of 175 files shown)


R/misc.R
```diff
@@ -859,9 +859,8 @@ check_new_data <- function(req, object, new_data) {
   step_cls <- class(object)[1]
   step_id <- object$id
   cli::cli_abort(
-    "The following required {cli::qty(col_diff)} column{?s} {?is/are} \
-    missing from `new_data` in step '{step_id}': {col_diff}.",
```
EmilHvitfeldt, line 863: The newer error chaining we are doing, makes it so we don't really have to mention the step name any more. so i did that small change as well

R/misc.R
```diff
@@ -859,9 +859,8 @@ check_new_data <- function(req, object, new_data) {
...
-    class = "new_data_missing_column",
+    "The following required {cli::qty(col_diff)} column{?s} {?is/are} missing 
+    from `new_data`: {col_diff}.",
```
hfrick, line 863:
```suggestion
    "The following required {cli::qty(col_diff)} column{?s} {?is/are} missing 
    from {.arg new_data}: {col_diff}.",
```

R/misc.R
```diff
@@ -859,9 +859,8 @@ check_new_data <- function(req, object, new_data) {
   step_cls <- class(object)[1]
   step_id <- object$id
   cli::cli_abort(
-    "The following required {cli::qty(col_diff)} column{?s} {?is/are} \
-    missing from `new_data` in step '{step_id}': {col_diff}.",
-    class = "new_data_missing_column",
```
hfrick, line 864: What was the reason for classing this error previously?

tests/testthat/test-colcheck.R
```diff
@@ -95,7 +95,10 @@ test_that("non-standard roles during bake/predict", {
 
   # This should require 'date' to predict.
   # The error comes from hardhat, so we don't snapshot it because we don't own it.
```
hfrick, line 97: I think this is a good principle to rely on, i.e. I would suggest not doing the change below and similar. If you want a more thorough test than "just make sure there is _a_ error" you could see if it's a classed error and check for that error class here.

tests/testthat/test-rename_at.R
```diff
@@ -40,7 +40,8 @@ test_that("mulitple functions", {
 
 test_that("no input", {
   # Wait for call pass through
```
hfrick, line 42: Is this still valid? if this refers to `step_mutate_at()` showing up in the error message, then we can remove this.

tests/testthat/test-window.R
```diff
@@ -20,7 +20,8 @@ test_that("error checks", {
     rec %>% step_window(y1, size = NA)
   )
   # Wait for call pass through
```
hfrick, line 22: same comment as on `step_mutate_at()` here

R/mutate_at.R
```diff
@@ -51,6 +51,10 @@ step_mutate_at <- function(recipe, ...,
                            inputs = NULL,
                            skip = FALSE,
                            id = rand_id("mutate_at")) {
+  if (rlang::is_missing(fn)) {
+    cli::cli_abort("Argument {.arg fn} must be specified.")
+  }
```
hfrick, line 56: nice one

hfrick (commented):
Nice one overall but I have questions/opinions on classed errors 🤓

R/misc.R
EmilHvitfeldt: this was from before `expect_snapshot()` so they were tested with `expect_error()`. Classed errors were used to make sure the correct error bubbled up. Instead of relying on regex matching.

However it is the only place we use it and only for testing purposes, and for that reason I think we are better off using the unclassed errors with `expect_snapshot()`

also, since we were matching on class, we never checked that the {cli} syntax was correct as we didn't see the error

i'm not opposed to classed errors. And they might even be useful if cases where {tune} could match on them.

tests/testthat/test-rename_at.R
EmilHvitfeldt: good eye! those are no longer needed

R/misc.R
hfrick: I'm okay with removing the error class if we are not using it. I see tests in the recipes extension packages that test for that error, but you probably know without checking if they test that a recipe step _from that extension package_ is the one throwing the classed error. 👍

tests/testthat/test-colcheck.R
hfrick: I'm unresolving this because it seems like you might have hit "resolved" by accident?

R/misc.R
EmilHvitfeldt: I ran a revdepcheck. it is only an issue for package i maintain so i think it is fine :) thanks for pointing it out

Issue #1358, opened 2024-08-12
`recipes:::strings2factors` should be restricted by the role


dajmcdon:
## Feature

`prep.recipe()` calls `recipes:::strings2factors()` to convert character columns to factors (when true). It would be better if this operated, not on all columns but only on those with certain roles (likely just predictors + outcomes).

Context: I'm working on an extension that would accommodate panel data. This involves keeping around a `key` column which needs to remain a character regardless of the decisions for predictor columns. 

I'm happy to do a PR for this if it would be helpful. I think it requires only minor changes, but hopefully would cause few downstream issues.

EmilHvitfeldt:
I think that sounds reasonable. If you feel like it, feel free to start the PR 😄

dajmcdon:
Closed by #1376

PR #1376, opened 2024-09-20
Strings2factors

dajmcdon:
This is a possible fix for #1358 .

* Adds a small function that undoes the results of `get_levels()` for any columns that do not have the roles "outcome" or "predictor".
* By "undoes", it simply gives those columns the same result that would happen for continuous variables `list(values = NA, ordered = NA)`

Sorry for the annoying whitespace changes. I think there's only a few.


R/misc.R
```diff
@@ -182,6 +182,18 @@ has_lvls <- function(info) {
   !vapply(info, function(x) all(is.na(x$values)), c(logic = TRUE))
 }
 
+kill_levels <- function(lvls, var_info) {
+  vars <- var_info$variable
+  roles <- var_info$role
+  preds_outcomes <- unique(vars[roles %in% c("outcome", "predictor")])
+  others <- unique(setdiff(vars, preds_outcomes))
+  if (length(others) > 0L) {
+    for (o in others) lvls[[o]] <- list(values = NA, ordered = NA)
```
EmilHvitfeldt, line 191:
```suggestion
    for (var in others) {
      lvls[[var]] <- list(values = NA, ordered = NA)
    }
```

EmilHvitfeldt (approved):
looking good so far. Can you also add a bullet to the news file and document this change in behavior for `prep.recipe()`? thank you

EmilHvitfeldt:
I also ran a revdepcheck on this PR and nothing bad happened

EmilHvitfeldt:
Looks good! thank you for doing this!

PR #1387, opened 2024-10-22
Add rlang type checkers

topepo:
Does some additional tests for basic things. The next step would be adding these checks to the steps.

PR #1391, opened 2024-11-08
Type checkers part 2

topepo:
Complement of #1387

This _should_ do it for the new type checkers in this package.

### DIFF
--- a/R/YeoJohnson.R
+++ b/R/YeoJohnson.R
@@ -111,7 +111,7 @@ prep.step_YeoJohnson <- function(x, training, info = NULL, ...) {
   check_type(training[, col_names], types = c("double", "integer"))
   check_number_whole(x$num_unique, args = "num_unique")
   check_bool(x$na_rm, arg = "na_rm")
-  if (!is.numeric(x$limits) | any(is.na(x$limits)) | length(x$limits) != 2) {
+  if (!is.numeric(x$limits) || any(is.na(x$limits)) || length(x$limits) != 2) {
     cli::cli_abort("{.arg limits} should be a numeric vector with two values,
                     not {.obj_type_friendly {x$limits}}")
   }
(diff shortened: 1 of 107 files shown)


R/other.R
```diff
@@ -107,13 +107,7 @@ step_other <-
            objects = NULL,
            skip = FALSE,
            id = rand_id("other")) {
-    if (!is_tune(threshold)) {
```
topepo, line 110: There's a lot of moving things here and in #1387. My rationale is that, if a the tuning process inserts a value, we would not check it. For example, `finalize_recipe()` does use the constructor. 

It does defer the checking until prep-time which is later than this but it seems like the only way to really be sure that we are getting corrrect inputs.

tests/testthat/test-window.R
```diff
@@ -14,35 +14,39 @@ rec <- recipe(~., data = sim_dat)
...
-    rec %>% step_window(y1, size = 2)
+    rec %>% step_window(y1, size = 2) %>% prep()
   )
   expect_snapshot(error = TRUE,
-    rec %>% step_window(y1, size = -1)
+    rec %>% step_window(y1, size = -1) %>% prep()
   )
   expect_snapshot(
-    rec %>% step_window(y1, size = pi)
+    rec %>% step_window(y1, size = 3 + .Machine$double.eps) %>% prep()
```
topepo, line 39: So I don't know that we need both of the tests using `.Machine$double.eps` but it did affirm by beleif of what _should_ happen given how `all_equal()` works.

R/YeoJohnson.R
```diff
@@ -109,6 +109,14 @@ step_YeoJohnson_new <-
 prep.step_YeoJohnson <- function(x, training, info = NULL, ...) {
   col_names <- recipes_eval_select(x$terms, training, info)
   check_type(training[, col_names], types = c("double", "integer"))
+  check_number_whole(x$num_unique, args = "num_unique")
+  check_bool(x$na_rm, arg = "na_rm")
+  if (!is.numeric(x$limits) | any(is.na(x$limits)) | length(x$limits) != 2) {
```
EmilHvitfeldt, line 114:
```suggestion
  if (!is.numeric(x$limits) || any(is.na(x$limits)) || length(x$limits) != 2) {
```

EmilHvitfeldt (approved):
amazing!

PR #1392, opened 2024-11-13
add `sparse` argument to `step_dummy()`

EmilHvitfeldt:
This PR adds the creation of sparse dummy variables in `step_dummy()` via the use of the `sparse` argument.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,8 @@
 
 * All steps and checks now require arguments `trained`, `skip`, `role`, and `id` at all times.
 
+* `step_dummy()` gained `sparse` argument. When set to `TRUE`, `step_dummy()` will produce sparse vectors.
+
 # recipes 1.1.0
 
 ## Improvements
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,7 +10,7 @@
 
 * All steps and checks now require arguments `trained`, `skip`, `role`, and `id` at all times.
 
-* `step_dummy()` gained `sparse` argument. When set to `TRUE`, `step_dummy()` will produce sparse vectors.
+* `step_dummy()` gained `sparse` argument. When set to `TRUE`, `step_dummy()` will produce sparse vectors. (#1392)
 
 # recipes 1.1.0
 
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -63,7 +63,8 @@
 #' this step.
 #'
 #' Also, there are a number of contrast methods that return fractional values.
-#' The columns returned by this step are doubles (not integers).
+#' The columns returned by this step are doubles (not integers) when 
+#' `sparse = FALSE`. The columns returned when `sparse = TRUE` are integers.
 #'
 #' The [package vignette for dummy variables](https://recipes.tidymodels.org/articles/Dummies.html)
 #' and interactions has more information.
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -18,6 +18,9 @@
 #' @param levels A list that contains the information needed to create dummy
 #'   variables for each variable contained in `terms`. This is `NULL` until the
 #'   step is trained by [prep()].
+#' @param sparse A logical. Should the columns produced be sparse vectors.
+#'   Sparsity is only supported for `"contr.treatment"` contrasts. Defaults to 
+#'   `FALSE`.
 #' @template step-return
 #' @family dummy variable and encoding steps
 #' @seealso [dummy_names()]
@@ -121,6 +124,7 @@ step_dummy <-
            preserve = deprecated(),
            naming = dummy_names,
            levels = NULL,
+           sparse = FALSE,
            keep_original_cols = FALSE,
            skip = FALSE,
            id = rand_id("dummy")) {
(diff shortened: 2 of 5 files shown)


EmilHvitfeldt:
it is set to default to `sparse` because it should only be used if you need it. BUT in workflows we plan to detect if the model accepts sparsity, then toggle sparsity in steps that can do it.

Example helper function below. Currently we don't have the infrastructure yet to determine if the sparse columns produced by `step_dummy()` will make it unharmed through. If `step_normalize()` was placed after `step_dummy()` in this example, the sparse dummies would be ruined. But we might be able to do so in the future. For that reason i think the helper should live in {recipes} instead of {workflows}. Any objections @simonpcouch?

``` r
library(recipes)
library(modeldata)

recipes_all_sparse <- function(x) {
  for (i in seq_along(x$steps)) {
    if (!is.null(x$steps[[i]]$sparse)) {
      x$steps[[i]]$sparse <- TRUE
    }
  }
  x
}

rec_spec <- recipe(Sale_Price ~ ., data = ames) |>
  step_unknown(all_nominal_predictors()) |>
  step_impute_mean(all_numeric_predictors()) |>
  step_normalize(all_numeric_predictors()) |>
  step_dummy(all_nominal_predictors())

rec_spec |>
  prep() |>
  bake(NULL) |>
  lobstr::obj_size()
#> 7.46 MB

rec_spec |>
  recipes_all_sparse() |>
  prep() |>
  bake(NULL) |>
  lobstr::obj_size()
#> 1.71 MB
```

<sup>Created on 2024-11-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

lattaai7:
Hello, I tried to solve the issue.

This is what I did:

This change adds the ability to create sparse dummy variables in step_dummy() by introducing a new 'sparse' argument. When set to TRUE, it will generate sparse dummy variables instead of dense ones, which can be more memory-efficient for datasets with many categorical levels.

You can review changes in this commit: https://github.com/lattaai7/tidymodels-recipes-1392/commit/74d4b46d54d943c2a542ee9bc4df73f8387f939d.


> [!CAUTION]
> Disclaimer: The commit was created by Latta AI and you should never copy paste this code before you check the correctness of generated code. Solution might not be complete, you should use this code as an inspiration only.

---
This issue was tried to solve for free by Latta AI - https://latta.ai/ourmission

If you no longer want Latta AI to attempt solving issues on your repository, you can block this account.

R/dummy.R
```diff
@@ -285,43 +293,60 @@ bake.step_dummy <- function(object, new_data, ...) {
...
-        new_data[[col_name]],
-        levels = levels_values,
-        ordered = is_ordered
-      )
+    if (object$sparse) {
+      current_contrast <- getOption("contrasts")[is_ordered + 1]
+      if (current_contrast != "contr.treatment") {
+        cli::cli_abort(
+          "when {.code sparse = TRUE}, only {.val contr.treatment} contrasts are
+          supported. Not {.val {current_contrast}}."
```
simonpcouch, line 308:
```suggestion
          "When {.code sparse = TRUE}, only {.val contr.treatment} contrasts are
          supported, not {.val {current_contrast}}."
```

tests/testthat/test-dummy.R
```diff
@@ -354,6 +354,34 @@ test_that("throws an informative error for single level", {
   )
 })
 
+test_that("sparse = TRUE works", {
+  rec <- recipe(~ ., data = tibble(x = c(NA, letters)))
+
+  suppressWarnings({
+    dense <- rec %>% step_dummy(x, sparse = FALSE) %>% prep() %>% bake(NULL)
+    dense <- purrr::map(dense, as.integer) %>% tibble::new_tibble()
```
simonpcouch, line 362: There's a portion in the `step_dummy()` documentation that reads "The columns returned by this step are doubles (not integers)"—let's clarify that that's only the case when `sparse = FALSE`.

NEWS.md
```diff
@@ -10,6 +10,8 @@
 
 * All steps and checks now require arguments `trained`, `skip`, `role`, and `id` at all times.
 
+* `step_dummy()` gained `sparse` argument. When set to `TRUE`, `step_dummy()` will produce sparse vectors.
```
simonpcouch, line 13:
```suggestion
* `step_dummy()` gained `sparse` argument. When set to `TRUE`, `step_dummy()` will produce sparse vectors. (#1392)
```

simonpcouch (approved):
Super clean!

No objections for that helper living in recipes rather than workflows.

Good move on scoping this PR as just implementing `step_dummy(sparse)` and holding off on tackling the "smart" toggling the argument in the backend. Your thoughts about what that interface might look like sound reasonable!

tests/testthat/test-dummy.R
EmilHvitfeldt: done in [`6e23367` (#1392)](https://github.com/tidymodels/recipes/pull/1392/commits/6e23367e089c82726b89791a205ffc28a5bd9749)

PR #1407, opened 2025-01-11
change sparse argument of step_dummy() to be an enum

### LINKED ISSUE #1396: have `sparse` argument in `step_dummy()` be a enum instead of a logical
this way it can be "auto", "yes", "no" 

this way, {workflows} has a way of knowing what should be turned overwritten and when

EmilHvitfeldt:
to close #1396

### DIFF
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -304,17 +304,19 @@ bake.step_dummy <- function(object, new_data, ...) {
 
     if (object$sparse == "yes") {
       current_contrast <- getOption("contrasts")[is_ordered + 1]
-      if (current_contrast != "contr.treatment") {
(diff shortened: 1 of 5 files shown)


R/dummy.R
```diff
@@ -301,7 +302,7 @@ bake.step_dummy <- function(object, new_data, ...) {
       ordered = is_ordered
     )
 
-    if (object$sparse) {
+    if (object$sparse == "yes") {
       current_contrast <- getOption("contrasts")[is_ordered + 1]
       if (current_contrast != "contr.treatment") {
```
topepo, line 307: Would it work with `contr_one_hot()`?

EmilHvitfeldt: correct, has been fixed

PR #1415, opened 2025-01-22
turn table to tibble in recipe.formula

EmilHvitfeldt:
This issue was found during revdepcheck: https://github.com/spsanderson/healthyR.ai/pull/349

we didn't use to check what type the data was that was passed to `recipe.formula()` which resulted in anything that passed `tibble::as_tibble()` worked. This is not ideal. This PR lets us release soon without having to fix it. Then we can later remove it so the functions does what the documentation says

### DIFF
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -209,6 +209,10 @@ recipe.formula <- function(formula, data, ...) {
     cli::cli_abort("{.arg data} is missing with no default.")
   }
 
+  if (is.table(data)) {
+    data <- as_tibble(data)
+  }
+
   if (!is.data.frame(data) && !is.matrix(data) && !is_sparse_matrix(data)) {
     cli::cli_abort(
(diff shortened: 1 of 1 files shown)


R/recipe.R
```diff
@@ -209,6 +209,10 @@ recipe.formula <- function(formula, data, ...) {
     cli::cli_abort("{.arg data} is missing with no default.")
   }
 
+  if (is.table(data)) {
```
topepo, line 212: Should we add a warning or deprecation notice for this? I'd like to get rid of this down the road; it shouldn't be supported.

topepo (commented):
Thanks for doing this and the PR in their repo.

R/recipe.R
EmilHvitfeldt: sure, i'm gonna add a warning and remove this in the following release with https://github.com/tidymodels/recipes/issues/1416

Issue #1455, opened 2025-03-26
`step_spline_*()` steps give uninformative errors when applied to constant variables


EmilHvitfeldt:
``` r
library(recipes)

mtcars$disp <- 3

recipe(.~ disp, data = mtcars) %>%
  step_spline_convex(disp) %>%
  prep()
#> Error in `step_spline_convex()`:
#> Caused by error in `spline_msg()`:
#> ! Error in eval(expr, envir) : Cannot set boundary knots from x.
```

<sup>Created on 2025-03-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #1457, opened 2025-03-26
add and use check_zv() in spline steps

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/1455

``` r
library(recipes)

mtcars$disp <- 3

recipe(.~ disp, data = mtcars) %>%
  step_spline_convex(disp) %>%
  prep()
#> Error in `step_spline_convex()`:
#> ! The following column has zero variance making computations unable to
#>   proceed: disp.
#> ℹ Consider using ?step_zv (`?recipes::step_zv()`) to remove those columns
#>   before this step.
```


EmilHvitfeldt:
wrote the check function to be useable in other steps as i'm sure this won't be the first time this is an issue

Issue #1296, opened 2024-04-03
`tune_args.recipe` doesn't include un-tunable arguments marked for tuning


simonpcouch:
Related to [tidymodels/tune#660](https://github.com/tidymodels/tune/issues/660#issuecomment-2035486249). It looks like, currently, `tune_args()` doesn't include arguments marked for tuning if they're not tunable. 

``` r
library(tidymodels)

rec <- 
  recipe(Price ~ ., data = credit_data) %>%
  step_impute_bag(
    Status, Home, Marital, Job, Income, Assets, Debt,
    # tunable
    trees = tune(),
    # not known tunable
    options = tune()
  )

credit_wflow <- workflow(rec, linear_reg())

# correct - we only have parameter info for trees
tunable(credit_wflow)
#> # A tibble: 1 × 5
#>   name  call_info        source component       component_id    
#>   <chr> <list>           <chr>  <chr>           <chr>           
#> 1 trees <named list [3]> recipe step_impute_bag impute_bag_EUEU7

# incorrect - doesn't include `options` from recipe
tune_args(credit_wflow)
#> # A tibble: 1 × 6
#>   name  tunable id    source component       component_id    
#>   <chr> <lgl>   <chr> <chr>  <chr>           <chr>           
#> 1 trees TRUE    trees recipe step_impute_bag impute_bag_EUEU7
```

<sup>Created on 2024-04-03 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

PR #1461, opened 2025-03-26
have tune_args.step return all arguments marked as tune

EmilHvitfeldt:
to close https://github.com/tidymodels/recipes/issues/1296

@simonpcouch i hope i understood the problem correctly

### DIFF
--- a/R/tune_args.R
+++ b/R/tune_args.R
@@ -32,9 +32,6 @@ tune_args.step <- function(object, full = FALSE, ...) {
 
   tune_param_list <- tunable(object)$name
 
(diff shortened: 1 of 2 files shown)


simonpcouch (approved):
This seems like the right fix for the issue locally to recipes! That said, tune relies on `tune_args()` quite heavily and in a way that I'd suspect would be sensitive to this change—likely worth revdepchecking this against tune + running extratests with this ref.

EmilHvitfeldt:
Ran revdepcheck and extratests and didn't see anything bad 🎉 
\

PR #1465, opened 2025-03-31
fix: remove unused line in yeo-johnson

### LINKED ISSUE #1464: Step Yeo Johnson contains an unused line
Hi, we're reusing some of your [step_YeoJohnson code in epipredict](https://github.com/cmu-delphi/epipredict/pull/451) and ran across an unused line of code in the YJ parameter estimation [here](https://github.com/tidymodels/recipes/blob/451b4df94d5352a2aaa33c131f2f1743edb414c5/R/YeoJohnson.R#L232). We looked into the original Yeo Johnson (2000) paper and found that indeed, mu_t is not needed for the estimation process, so we removed the line in our extension without issue. Figured we'd upstream the change.

dshemetov:
Close #1464 .


EmilHvitfeldt:
Very good catch! thank you for the issue and PR to fix it

Issue #1349, opened 2024-07-12
`step_dummy()` doesn't capture contrasts


EmilHvitfeldt:
`step_dummy()` doesn't capture contrasts, thus won't work `bake()` if contrasts option if different between `prep()` and `bake()` time.

``` r
library(recipes)

rec_normal <- recipe(~ ., data = iris) %>%
  step_dummy(all_nominal_predictors()) %>%
  prep()

param <- getOption("contrasts")
go_helmert <- param
go_helmert["unordered"] <- "contr.helmert"
options(contrasts = go_helmert)

rec_helmert <- recipe(~ ., data = iris) %>%
  step_dummy(all_nominal_predictors()) %>%
  prep()

rec_helmert %>%
  bake(iris)
#> # A tibble: 150 × 6
#>    Sepal.Length Sepal.Width Petal.Length Petal.Width Species_X1 Species_X2
#>           <dbl>       <dbl>        <dbl>       <dbl>      <dbl>      <dbl>
#>  1          5.1         3.5          1.4         0.2         -1         -1
#>  2          4.9         3            1.4         0.2         -1         -1
#>  3          4.7         3.2          1.3         0.2         -1         -1
#>  4          4.6         3.1          1.5         0.2         -1         -1
#>  5          5           3.6          1.4         0.2         -1         -1
#>  6          5.4         3.9          1.7         0.4         -1         -1
#>  7          4.6         3.4          1.4         0.3         -1         -1
#>  8          5           3.4          1.5         0.2         -1         -1
#>  9          4.4         2.9          1.4         0.2         -1         -1
#> 10          4.9         3.1          1.5         0.1         -1         -1
#> # ℹ 140 more rows
```

Everything works up and until now. But `rec_normal()` was prepped using `contr.treatment` so it breaks when used in `contr.helmet`

``` r

rec_normal %>%
  bake(iris)
#> Error in `vctrs::vec_locate_matches()`:
#> ! Each value of `needles` must have a match in `haystack`.
#> ✖ Location 5 of `needles` does not have a match.
```

I could imagine that this will break some production applications, where prepping was done using `contr.poly`, but the prediction environment (vetiver as an example) would have no idea.

# Suggested fix

I think this bug give enough motivation to add a contrast argument to `step_dummy()` such that:

- it can keep the information
- simplify the code a little
- long term: start deprecation of using global options in steps

EmilHvitfeldt:
Adding `contrast` argument

[...]

PR #1484, opened 2025-04-05
add `contrasts` argument to `step_dummy()`

EmilHvitfeldt:
To close https://github.com/tidymodels/recipes/issues/1349

This PR:

- Adds `contrasts` argument to `step_dummy()`. It takes a single contrast function or a a list of contrasts functions as input. I intentionally didn't allow string representation `c(unordered = "contr.treatment)` as it feels messy AND it would be annoying for parallel processing.
- Added a deprecation warning if someone sets `option(contrasts)` (by checking if it is different that default) to steer them towards using the explicit argument.
- Updated documentation to to reflect this both in function and dummies vignette.

## Revdepcheck

Ran clean

## Books/websites

Not used: https://github.com/search?q=org:tidymodels+getOption+contrasts&type=code

### DIFF
--- a/R/dummy.R
+++ b/R/dummy.R
@@ -232,6 +232,22 @@ prep.step_dummy <- function(x, training, info = NULL, ...) {
 
   check_contrasts_arg(x$contrasts)
 
+  if (
+    !identical(
+      getOption("contrasts"),
+      c(unordered = "contr.treatment", ordered = "contr.poly")
+    )
+  ) {
+    x$contrasts <- getOption("contrasts")
+    x$contrasts <- lapply(x$contrasts, get)
+
+    lifecycle::deprecate_warn(
+      "1.3.0",
+      I("options(contrasts) with step_dummy()"),
+      I("step_dummy(contrasts)")
+    )
+  }
+
   if (length(col_names) > 0) {
     ## I hate doing this but currently we are going to have
     ## to save the terms object from the original (= training)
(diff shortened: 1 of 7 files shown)


R/dummy.R
```diff
@@ -98,35 +100,50 @@
...
+#' tidy(dummies_helmert, number = 1)
 step_dummy <-
   function(
     recipe,
     ...,
     role = "predictor",
     trained = FALSE,
     one_hot = FALSE,
+    contrasts = list(
+      unordered = stats::contr.treatment,
```
topepo, line 144: I think that using the function value would attach additional environments to the object (e.g. packages and global). I'm ok with character values but maybe require a call instead (or convert these to a call?).

tests/testthat/test-dummy.R
```diff
@@ -151,6 +151,137 @@ test_that("create all dummy variables", {
   )
 })
 
+test_that("make sure contrasts argument work", {
```
topepo, line 154: Can you try an example with `hardhat::contr_one_hot()` or some other external contrast function?

topepo (commented):
Looks good but let's think about different input values (to avoid attaching environments)

R/dummy.R
EmilHvitfeldt: done in https://github.com/tidymodels/recipes/pull/1484/commits/ad7d62d6321bc82265b79644381fd8253f62ce0e

tests/testthat/test-dummy.R
EmilHvitfeldt: done in https://github.com/tidymodels/recipes/pull/1484/commits/f49dac1229d4c982339e542211beefca179b1851

PR #1544, opened 2026-05-29
move mixOmics to suggests

EmilHvitfeldt:


mattmalcher:
ty for this - it was causing renv to treat recipes as a bioconductor package.

PR #1552, opened 2026-08-20
Add stop_at argument to bake()

### LINKED ISSUE #1551: stage-wise aplication of a recipe
@dicook has a request to be able to apply a subset of recipe steps (details in [this presentation](https://dicook.github.io/useR-2026/slides.html#/better-integration-with-visualisation-for-diagnostics); pre-processing tab): 

```r
rec <- recipe(outcome ~ ., data = train) |>
  step_BoxCox(all_numeric_predictors()) |>
  step_other(all_nominal_predictors(), threshold = .1) 

# Extract the data as it exists after each step
rec |>
  prep() |>
  taste(after = c(1,2))
```

I suspect that we can easily add an option to `bake()` for this.

EmilHvitfeldt:
Closes #1551.

Adds a `stop_at` argument to `bake()` so users can see the data as it exists partway through a prepped recipe, which is handy for visual diagnostics.

```r
rec <- recipe(cyl ~ ., mtcars) |>
  step_center(all_predictors()) |>
  step_pca(all_predictors(), num_comp = 2) |>
  prep()

bake(rec, mtcars, stop_at = 1)              # by step number
bake(rec, mtcars, stop_at = "center_zo5Cx") # by step id
```

Details:

* `stop_at` sits after `...`, so it is named-only and can't be confused with tidyselect terms. The `bake()` generic is `function(object, ...)`, so it needed no change. Default `NULL` keeps the current behavior.
* Steps with `skip = TRUE` keep their existing semantics and still count toward the numbering, so numbers line up with `tidy(rec)`.
* `stop_at` errors with `new_data = NULL`. That path returns `object$template`, the fully processed training set, and `prep()` does not retain the raw training data, so intermediate stages can't be reconstructed. The message points users at passing the training data to `new_data`.
* The post-loop column selection normally uses `last_term_info`, which describes the data after *all* steps and so is wrong for a partial bake. Deriving the info from the intermediate data alone wasn't enough: roles for derived columns were lost, and `all_numeric_predictors()` selected nothing after `step_pca()`. Instead `update_term_info()` tracks the term info as each step is applied, mirroring the logic in `prep()`. It only runs when `stop_at` is supplied, so the default path is unaffected.

Tests cover step numbers, step ids, selectors, `composition = "dgCMatrix"`, a mid-recipe `skip = TRUE` step, and snapshot errors for out-of-range numbers, non-whole numbers, unknown ids, wrong types, length > 1, `new_data = NULL`, and a recipe with no steps.

`devtools::check()` is clean: 0 errors, 0 warnings, 0 notes.

### DIFF
--- a/R/recipe.R
+++ b/R/recipe.R
@@ -892,16 +892,22 @@ resolve_stop_at <- function(stop_at, object, call = caller_env()) {
     )
   }
 
-  if (is.character(stop_at)) {
-    check_string(stop_at, arg = "stop_at", call = call)
+  if (length(stop_at) != 1 || is.na(stop_at)) {
+    cli::cli_abort(
+      "{.arg stop_at} must be a single step number or step id, \\
+      not {.obj_type_friendly {stop_at}}.",
+      call = call
+    )
+  }
 
+  if (is.character(stop_at)) {
     ids <- map_chr(object$steps, \(x) x$id)
     loc <- which(ids == stop_at)
 
     if (length(loc) == 0) {
       cli::cli_abort(
         c(
-          "x" = "{.arg stop_at} must be a step number or step id, \\
+          "x" = "{.arg stop_at} must be a single step number or step id, \\
                  not {.val {stop_at}}.",
           "i" = "The step ids of {.arg object} are {.or {.val {ids}}}."
         ),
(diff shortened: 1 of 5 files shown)


topepo:
@dicook how does this look?

R/recipe.R
```diff
@@ -782,7 +813,16 @@ bake.recipe <- function(object, new_data, ..., composition = "tibble") {
   bakeable_names <- intersect(original_training_names, original_names)
   new_data <- new_data[, bakeable_names]
 
-  n_steps <- length(object$steps)
+  n_steps <- resolve_stop_at(stop_at, object)
+
+  # `last_term_info` describes the data after *all* steps, so it cannot be used
```
topepo, line 818: I wouldn't have thought of that. Nice

R/recipe.R
```diff
@@ -835,6 +878,95 @@ bake.recipe <- function(object, new_data, ..., composition = "tibble") {
...
+
+  if (!is.finite(stop_at) || stop_at != trunc(stop_at)) {
+    cli::cli_abort(
+      "{.arg stop_at} must be a whole step number, not {stop_at}.",
+      call = call
+    )
+  }
+
+  if (stop_at < 1 || stop_at > n_steps) {
+    allowed <- if (n_steps == 1) "1" else paste0("between 1 and ", n_steps)
```
topepo: WAAAAAT

Was this a test? 😼

topepo (changes requested):

R/recipe.R
EmilHvitfeldt: no one saw this

topepo (approved):

dicook:
Thanks @topepo I'm just looking at this now. I'll give it a run-through this next week.

Issue #1567, opened 2026-09-01
`step_relevel()` reports the wrong columns in its reference-level error


EmilHvitfeldt:
`prep.step_relevel()` (`R/relevel.R`) computes `order_check` (whether each column is an ordered factor) and `ref_check` (whether the reference level is missing from each column), but then uses `order_check` to build the error message's offending-column list instead of `ref_check`:

```r
order_check <- map_lgl(objects, attr, "is_ordered")
ref_check <- map_lgl(objects, function(x, y) !y %in% x, y = x$ref_level)
if (any(ref_check)) {
  offenders <- names(order_check)[!order_check]
  cli::cli_abort(
    "...doesn't include required reference level {.val {x$ref_level}}: {.var {offenders}}."
  )
}
```

The condition correctly triggers on `ref_check`, but `offenders` is derived from `!order_check`, so the reported column names have nothing to do with which columns actually lack the reference level.

Suggested fix: change `offenders <- names(order_check)[!order_check]` to `offenders <- names(ref_check)[ref_check]`.

PR #1572, opened 2026-09-17
Fix step_relevel() reporting wrong columns in reference-level error (#1567)

luisDVA:
tidy dev day 2026
prep.step_relevel() built the offending-column list from order_check (ordered-factor status) instead of ref_check (missing reference level), so the error named columns unrelated to the actual problem.  

addresses #1567


topepo:
The error in formatting is not a big issue.

topepo (approved):

